PR #2, opened 2018-01-09
Update Algo.h

kasome:
Try to speed up a little for the for loop in these functions: FindMatch, FindIndex, FindAll


klugier:
kasome - it is nice to see that you made first pull request with real change to ultimatepp github mirror. Currently we are still on SVN, so pull request is impossible.

However, I informed Mirek about this request, so it should be process normally.

If you want to push your fixes in that way, it is fine for me. If you want more advanced tools like access to our redmine or Slack - please let me know.

Have a nice day!

PR #8, opened 2019-11-18
Core: IsDoubleWidth() function to simplify handling of double width Unicode characters

ismail-yilmaz:
I propose adding IsDoubleWidth() function to Upp/Core.
It simplifies Eastern-Asian/CJK character handling, and seems to cover more than IsCJKIdeograph().
This code is a "rip-off" from [Markus Kuhns wcwidth() implementation.](https://www.cl.cam.ac.uk/~mgk25/ucs/wcwidth.c). It is widely used, and doesn't seem to have a problematic license.


mirek-fidler:
Thanks, applied

PR #11, opened 2020-02-12
Don't pull in libXtst on POSIX when NO_XTEST is set

erikjwaxx:
Currently, the SysInfo package will pull in libXtst on POSIX systems regardless of whether or not NO_XTEST is set at build time.


izabala123:
I am sorry erikjwaxx. I do not know what is NO_XTEST.

PR #12, opened 2020-03-03
VecLine::Clip infinite recursion fix

erikjwaxx:
It looks like, from looking at the function signature `Clip(const Rectf& rect) const` that the intent here is to keep the operated-on `VecLine` immutable and return a _copy_ clipped by `rect`.

This change effects that logic while breaking the infinite recursion.


mirek-fidler:
Thanks, applied. That said, whole Geom package is sort of deprecated (now properly marked as such in description)

PR #13, opened 2020-03-05
SysInfo: fix MouseKey.cpp compile w/ NO_XTEST on non-Win32 platforms

erikjwaxx:
The `#if defined(PLATFORM_WIN32) || !defined(flagNO_XTEST)` directive
was incorrectly `#endif`'d outside the `namespace Upp` block it was opened
in, leading to a syntactically invalid file with an unterminated
namespace block when this `#ifdef` is false.


klugier:
@erikjwaxx - thanks already on trunk. Merged by @izabala123.

PR #16, opened 2020-05-19
FUNDING.yml

klugier:
This simply change adds "Sponsorship" button that displays link to the U++ donation site. Current;y, In the settings I enable sponsorship button, however it needs FUNDING.yml file to work correctly.


mirek-fidler:
I do not like the idea of adding .github into upp root.

PR #17, opened 2020-05-19
Travis CI integration

klugier:
Travis CI integrated for following build platforms:
- AMD64 - Clang
- AMD64 - GCC
- ARM64 - GCC

Travis CI is free for open source projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,11 @@ matrix:
       osx_image: xcode11.4
 
 install:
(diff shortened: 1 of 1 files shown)


klugier:
I think we should compile using UMK. Generated script will not be optimal in case when files are modified on separate branch.

ped7g:
btw why Travis? I have positive experience with Cirrus CI. Should I try to create PR with cirrus config so you can evaluate?

PR #21, opened 2020-09-08
CtrlCore: Removing Turtle specific flags

ismail-yilmaz:
This pull request aims to remove the now obsolete TURTLE flags defined in CtrlCore.
The new Turtle package does not require these flags. It relies on VIRTUALGUI flag.

Note that there is one other TURTLE flag yet to be removed (or changed to VIRTUALGUI) (in Core/App.cpp):

https://github.com/ultimatepp/ultimatepp/blob/0b2bd0a24170d6b8af1cacc027fd3911f006fcb9/uppsrc/Core/App.cpp#L540-L549

This needs to be handled carefully as it redefines Upp::LaunchWebBrowser global function on TURTLE backend.
(I think we should simply add a global WhenLaunchWebBrowser event in VIRTUALGUI: backend, and let clients handle it as they see fit.


klugier:
@mirek-fidler is this something we could merge. It seems that it is pending for the long time.

@ismail-yilmaz is it still valid?

ismail-yilmaz:
@klugier 

It seems that I've messed this up while syncing. I'm closing this, sorry. Changes are simple. We can do it later.

PR #23, opened 2020-10-08
a few fixes/additions to plugin/zip

70MM13:
- clear the Vector<File> file before reading (new) directory list
- add default constructor to FileUnzip (for convience)
- add Close() method to Unzip (tidy up after use)


mirek-fidler:
Thanks applied.

PR #24, opened 2020-10-14
README changes.

ismail-yilmaz:
README file does not list Clang in supported compilers list.

This PR:
- Adds "Clang" entry to supported compilers list.
- Further  smplifies the example GUI code.


klugier:
Agreed that we need to add clang. Not sure about changes in example - the current version is fine and show some functionalities like setting the size of the window. "class MyApp: public Upp::TopWindow {" this is Window indeed, so the MyAppWindow is more verbose and show what the class really is.

Formatting - we do not have formatting guild-lines inside upp, so the current version is fine. So generally speaking changes in sample code is a slippery subject :) I will keep current version without any modificaiton.

ismail-yilmaz:
> Not sure about changes in example - the current version is fine and show some functionalities like setting the size of the window. "class MyApp: public Upp::TopWindow {" this is Window indeed, so the MyAppWindow is more verbose and show what the class really is.

No problem. :)
That said, the main function of this example GUI app is to show the simplicity and initial coding patterns for U++. And my rationale is that the proposed version is closer to the CtrlLib project templates and a single line in main loop might give a more positive first impression to newbies. Not to mention that it is obvious from the proposed code that the window geometry can be changed using the SetRect method. Thus,  IMO nothing seems to be lost in the proposed code. I admit that I am bad at naming things though. :D

klugier:
Now I see SetRect was moved to the constructor. You are right about CtrLib approch - just checked it.

So, as you see right now I do not have any objections :)

mirek-fidler:
Thanks applied.

PR #55, opened 2021-12-22
CtrlLib: Gui14 tutorial example now sets minimal window size in order…

klugier:
In order to correctly render Gui14 example content the window should posses minimal size. The fix improve the example by adding SetMinSize.

### DIFF
--- a/tutorial/Gui14/main.cpp
+++ b/tutorial/Gui14/main.cpp
@@ -6,7 +6,7 @@ struct MyAppWindow : TopWindow {
 	Button lt, rt, lb, rb, lv, ht, hv, cb, rc;
 
 	MyAppWindow() {
-		Title("My application with button").Sizeable();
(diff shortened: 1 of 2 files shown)


mirek-fidler:
IDK, in production code definitely, but this is example. And maybe showing what happens if you do not have minsize defined there is helpful as well? It is not like it crashes or behaves unpredicatably...

klugier:
I was thinking more about introducing SetMinSize into tutorial. This function is not present in the GUI tutorial and I think it is the important one. The only place when it is explicitly attached is in code generated by layout manager. I think it is not enough. Anyway, if you think it shouldn't be present just close the PR otherwise just merge it...

Also, I do not exactly remembered, but somebody on some forum just raised the issue that the window in this tutorial point behaves strangely.

mirek-fidler:
OK, I have added 14.1 specifically designed to showcase the problem and SetMinSize. Please check.

PR #56, opened 2021-12-23
Instant Setup for VS Professional 2022

klugier:
What's changed:
- Added detection for Visual Studio Professional 2022 detection
- The configuration is now gather from both x86 and x64 directories (user could have both)
- minor code reorganization


mirek-fidler:
Thanks, very helpful.

PR #59, opened 2021-12-30
ScrollBar rewritten.

Lance219:
`CtrlLib/ScrollBar` is modified to get rid of the `prev, prev2, next, next3` Buttons. The corresponding portion of the `ScrollBar` are painted used the similar logic that original code used to paint the 3 parts of Slider&Thumb. The code that's changed mainly involve painting and mouse event. No interface is changed, so the revision is transparent to libary users.

As a result of the revision. Now `sizeof(ScrollBar)==sizeof(Button)`. Initial tests indicate that executable size also benefit from the revision.


uppsrc/CtrlCore/CtrlCore.h
```diff
@@ -481,7 +481,9 @@ class Ctrl : public Pte<Ctrl> {
 	};
 
 	Top         *top;
-	int          exitcode;
+#if defined(_M_IX86) // 32bit MSC compiler
+	int			exitcode;
+#endif
```
klugier, line 486: Well - it can not be like this. What is the reason of making this code dedicated to _M_IX86 platform?

uppsrc/CtrlLib/ScrollBar.cpp
```diff
@@ -73,123 +58,97 @@ ScrollBar::ScrollBar() {
...
-	int dummy;
-	return Slider(dummy);
-}
-
-int& ScrollBar::HV(int& h, int& v) const
-{
-	return IsHorz() ? h : v;
-}
+	int  s0, s1, s2, s3, s5, // width/height of each section (s6==s0)
+		 total, m=3;
```
klugier, line 82: What is the reason of introducing new sections? Can not we just focus at the beginning of getting ride of Buttons that are keep on stack? I mean iterative approach to development. Decrees memory consumption in chunks - not everything at once. 

In this PR we are changing a lot. In my opinion too many things. Merging this might be very risky.

mirek-fidler:
This is interesting job. Not sure it solves any real problem though... How many Scrollbars do you have in the application?

That said, I will definitely try to reorganize Ctrl variables to reduce the footprint. String info is also probably ripe for optimisation (I guess raw char * would work just as well there with some effort in Ctrl).

uppsrc/CtrlCore/CtrlCore.h
Lance219: Klugier:

I am not very familiar with UPP compiler flags. Mirek should be able to come up with the more appropriate one. Theoretically (as far as I know), moving `exitcode` to new location to eliminate padding for 64bit platform (which works for CLANG64 and MSC 64) will not affect size of `Ctrl` on 32 bit platforms. Unfortunately tests indicate for MSBT22(32 bit), size of `Ctrl` will be increased by 8 bytes because of relocation of `exitcode`. This is undesirable. The _M_IX86 flags is used to work around the problem with MSBT22 (and microsoft complier in general). If mirek can use instead the flag the UPP used to differentiate 32bit and 64 bit platform, it will be ideal. `Ctrl` has been tuned perfectly for 32bit platform, we should leave it untouched to avoid introducing unexpected side effects.

uppsrc/CtrlLib/ScrollBar.cpp
Lance219: As `prev, prev2, next, next2` are removed, `ScrollBar::paint(...), ScrollBar::LeftDown(...), ScrollBar::LeftRepeat`, etc., are required to fake/mock what the individual buttons used to do as contained `Ctrl`s. In original code, `Slider` is divided into 3 parts (eg, in `GetMousePart(...)`): the upper blank erea, thumb, lower blank area. As we now need to paint the space allocated to fake `prev,next, prev2, next2` buttons, 4 more sections are added to create a total of 7 sections.

Lance219:
> 
mirek:

I know the saving on memory is mostly psychological and is not likely to serve much practical purposes.

That being said, I have performed a test on `uppsrc/ide` , with some static member variables to keep track of number of current instances of `ScrollBar` and maximum number of it that's used throughout the life of the program. These variables are updated in `ScrollBar` constructor and destructor. My test indicates **theide** easily use around 2000(1999 to be more specific) `ScrollBar`s. To repeat my test result, just open the **uppsrc/ide/ide.iml`** file in the ide that's created by the ide that's being debugged. To save time, I have uploaded the revised <CtrlLib/ScrollBar.*>  that keeps track of maximum number of ScrollBar instances used in a program's life span.

Not many programs will be as complex as **theide**, for sure. And even in this case, around 2M memory is nothing in todays hardware. So it's total up to you to decide whether you want to go ahead. 

BTW, the code actually simplified painting & point resolution logic, resulting in more compact executable.

Thanks,
Lance

ps: in another test, the maximum scrollbar logged go as high as 4100 (with `uppsrc/ide` being tested again). I debug run (F5) it and in the new ide that's created in last step, first I opened a skeleton CtrlLib project, in it I insert an empty abc.lay (so that the layout designer is brought out). Then I switch main projgect to `uppsrc/ide` in it and scroll to the very bottom and open `uppsrc/ide/ide.lay`, this time the maximum logged went up to 4100.

PR #60, opened 2022-01-07
File Sel's file list should use no round size mode

klugier:
Right now, the FileSel scaling behaves in a strange way. To overcome this issue NoRoundSize mode for file list should be enable.


mirek-fidler:
IDK, for me it behaved strange before NoRoundSize... that is why it is there right?

PR #77, opened 2022-06-18
Use dark window titlebar on Windows 10+ with dark themes

EricKotato:
This PR makes titlebar on Windows 10+ to match application dark mode.

Here's "Button" example compiled before and after applying the patch:

Compiled apps will also work on older systems that doesn't support attribute or even DWM itself.


klugier:
I created improved version of this PR in #80. The spirit is the same and this change is definitely needed to better support dark mode on Windows 10 and Windows 11. Thank you for your contribution!

PR #78, opened 2022-07-02
GridCtrl layout control has been move to Upp namespace.

klugier:
Changes:
- DropGrid.usc is now under Upp namespace
- GridCtrl.usc is now under Upp namespace
- GridCtrl references and examples has been ported to new namespace


mirek-fidler:
Thanks, good job.

PR #82, opened 2022-07-15
SSH: Deprecated methods are removed.

ismail-yilmaz:
Some deprecated functions were marked for removal in U++ 2021.
This pull requests removes those methods.

### DIFF
--- a/uppsrc/Core/SSH/Session.cpp
+++ b/uppsrc/Core/SSH/Session.cpp
@@ -205,30 +205,10 @@ bool SshSession::Connect(const String& host, int port, const String& user, const
 			if(!rc) {
 				LLOG("Handshake successful.");
 				WhenPhase(PHASE_AUTHORIZATION);
+				if(WhenVerify && !WhenVerify(host, port))
+					ThrowError(-1);
 			}
 			return !rc;
-	})) goto Bailout;
-
-	if(!Run([=] () mutable {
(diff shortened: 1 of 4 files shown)


uppsrc/Core/SSH/Session.cpp
```diff
@@ -205,32 +205,12 @@ bool SshSession::Connect(const String& host, int port, const String& user, const
 			if(!rc) {
 				LLOG("Handshake successful.");
 				WhenPhase(PHASE_AUTHORIZATION);
+				if(WhenVerify && !WhenVerify(host, port))
+					ThrowError(-1);
```
klugier, line 209: ThrowError(-1) is enigmatic. It should ThrowError with information...

ismail-yilmaz: @klugier 

It is a *generic failure* or rejection, separated from the libssh2's own error codes. Since the WhenVerify() hands the operation over to the user. (client), client code *should* deal with it (i.e. it can print a cancellation message or pop up an error dialog.)

ismail-yilmaz: Besides, setting the error code to -1, first checks for errors on libssh2 subsystem. (So it will print an error message if the failure is pertaining to the ssh subsystem.)

klugier (approved):
Looks good to me, however you should wait for Mirek's approval!

PR #89, opened 2022-08-23
Custom Dpi for PdfDraw

M0untacir:
issue:
DrawImage of an image with different DPI than 600 causes wrong pdf size output.


mirek-fidler:
Does not make much sense to me. Could you perhaps provide an example where PdfDraw misbehaves?

(As far as I know, Image DPI is completely ignored about everywhere in PdfDraw...)

M0untacir:
I made an example in the forum, with more explanation.
[https://www.ultimatepp.org/forums/index.php?t=msg&th=11891&start=0&](https://www.ultimatepp.org/forums/index.php?t=msg&th=11891&start=0&)

PR #90, opened 2022-09-03
FileSelNative::Execute() fix: default filename was some trash on saving

zsdebre:
[BUG + FIX: FileSelNative shows default filename as some trash characters](https://www.ultimatepp.org/forums/index.php?t=msg&th=11901&goto=58798&#msg_58798) [[message #58798](https://www.ultimatepp.org/forums/index.php?t=msg&th=11901&goto=58798&#msg_58798)]

### DIFF
--- a/uppsrc/CtrlLib/Win32.cpp
+++ b/uppsrc/CtrlLib/Win32.cpp
@@ -209,12 +209,11 @@ bool FileSelNative::Execute(bool open, const char *dlgtitle) {
 						out.Append( ToSystemCharsetW('\"' + fn + '\"'));
 					else
 						out.Append(ToSystemCharsetW(fn));
-					out.Add(0);
 				}
 			}
 		}
 		int l = min(out.GetCount()*2, bufsize - 1);
-		memcpy(buffer, out, l + 1);
+		memcpy(buffer, out, l);
 	}
 
 	if(dlgtitle)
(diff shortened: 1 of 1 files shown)


uppsrc/CtrlLib/Win32.cpp
```diff
@@ -206,14 +206,14 @@ bool FileSelNative::Execute(bool open, const char *dlgtitle) {
...
-						out << W32(fn);
-					out.Cat(0);
+						out.Append(ToSystemCharsetW(fn));
+					out.Add(0);
 				}
 			}
 		}
-		int l = min(out.GetLength(), bufsize - 1);
+		int l = min(out.GetCount()*2, bufsize - 1);
 		memcpy(buffer, out, l + 1);
```
mirek-fidler: I feel like "l + 1" here might be wrong, in fact the whole handling of terminal double \0 does not feel right...

zsdebre: Yes, you are absolutely right, thanks for your feelings :)
The "+ 1" is not needed, nor the "out.Add(0);" line.
I fixed in the branch.

mirek-fidler:
Uhm, those NULL characters added to the list are for multiple filenames.

https://docs.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew

You are supposed to separate files with NULL, then end the list with double NULL...

mirek-fidler:
Fixed with slightly modified code...

zsdebre:
Thanks!

PR #98, opened 2022-09-26
Fix for blank dialog icons in standard dialogs on GTK backend

klugier:
Since GTK version 3.10 icons names for dialog elements such as information, warning, error and exclamations has been deprecated. On newest version our current loading approach doesn't work at all. To eliminate the issue we need new theme string for GTK version 3.10 and above.

What's new:
- dedicated code for loading dialog icon for GTK version 3.10 and above
- more secure approach of loading GTK icons, if it fails stock icons are used instead of null one

We can remove if statment completly if we do not want to support GTK priori to version 3.10. 3.10 version was release on 2013-09-23. There is very little chances that some distribution are using it even with long term support.

Information about deprecation can be find [here](https://docs.gtk.org/gtk3/const.STOCK_DIALOG_INFO.html).

### DIFF
--- a/uppsrc/CtrlLib/ChGtk3.cpp
+++ b/uppsrc/CtrlLib/ChGtk3.cpp
@@ -280,6 +280,16 @@ Image Gtk_IconAdjusted(const char *icon_name, int size)
 	return m;
 }
 
+void Gtk_OverrideCtrlImgIcon(int i, const char *s)
+{
+	auto icon = Gtk_Icon(s, DPI(48));
+	if (icon.IsEmpty()) {
+		return;
+	}
+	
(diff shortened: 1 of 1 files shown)


uppsrc/CtrlLib/ChGtk3.cpp
```diff
@@ -314,11 +324,6 @@ void ChHostSkin()
 
 	ChBaseSkin();
 
-#if 0 // TODO (?)
-		{ SColorLight_Write, 2*5 + 0 },
-		{ SColorShadow_Write, 3*5 + 0 },
-#endif
```
klugier, line 320: Garbage...

PR #99, opened 2022-09-26
Core/Stream: GetUtf8() method now returns failure on overlong and invalid sequences.

ismail-yilmaz:
UTF-8 overlong sequences are considered a potential security threat.

For some reason, Stream::GetUtf8() methods allow up to 6 bytes sequences, which are invalid UTF-8.

This patch aims to improve the situation by returning failure (-1) also on overlong and invalid sequences.

Actual code is based on Upp::FetchUtf8() function, which also has better range checking and is faster..

Please review.


mirek-fidler:
If you have some testing code for this, could you provide a test for autotest as well?

ismail-yilmaz:
Will do.

ismail-yilmaz:
I've added an autotest for Stream::GetUTF8() method, based on Marcus Kuhn's [UTF-8 decoder capability and stress test](https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txtl).

This patch checks for malformed, overlong and invalid UTF-8 sequences. And it uses the *recommended* method, 
by giving the developer a chance to replace each unexpected byte with the replacement character (by not consuming them).

For example, a malformed two-byte UTF-8 sequence shall be replaced with two replacement characters (this allows easy debugging etc.) This is done by rewinding using Seek().

This gives near-exact results with industry standard apps' behavior on malformed UTF-8) with no known disadvantage.

Please check.

mirek-fidler:
Upon more refined testing, I have found one glitch:

https://stackoverflow.com/questions/6907297/can-utf-8-contain-zero-byte

mirek-fidler:
Another problem - that "rewind" Seek would only work on streams that actually can seek...

mirek-fidler:
Please review my fixes....

ismail-yilmaz:
> Upon more refined testing, I have found one glitch:

AFAIK, one place the overlong NUL is used is in Java, as it uses the so-called *modified UTF-8*. We can make an exception for the two bytes representation of NUL (`C0 80`), for the sake of compatibility.

> Another problem - that "rewind" Seek would only work on streams that actually can seek...

Can't we just check the lower bounds of the buffer and move the pointer back (ptr/pos), when possible?

> Please review my fixes....

Sorry, couldn't find the fixes. I only see my patch. Where are they?

mirek-fidler:
Forgot to push, sorry.

https://github.com/ultimatepp/ultimatepp/commit/6cd5e1aa280ced879863e284030bec70eece8751

Mirek

ismail-yilmaz:
@mirek-fidler 

OK, removing the Seek() won't really hurt, I guess.

1) The `pos` variable should be removed. (it is unused now.)
https://github.com/ultimatepp/ultimatepp/blob/6cd5e1aa280ced879863e284030bec70eece8751/uppsrc/Core/Stream.cpp#L237

2) As as side note: I thought this error was used to signal the unexpected end of data chunk.
https://github.com/ultimatepp/ultimatepp/blob/6cd5e1aa280ced879863e284030bec70eece8751/uppsrc/Core/Stream.cpp#L266

Seems fine.

Thanks for merging this. I'll have time to test this thoroughly in a working environment..

mirek-fidler:
LoadError is a general "input stream is broken" error. The most often used
with serialization, when bad input data is detected, usually with activated
exceptions. I think its use here is appropriate and, besides, consistent
with previous behaviour.

PR #103, opened 2022-10-01
A function to detect strings with incomplete (split) UTF8 bytes at the end.

ismail-yilmaz:
This patch adds a CheckUtf8Split() function (and its overloads) to Core/Utf8 helper functions, it also adds an autotest & Topic++ entry.

What  it does:

- Checks for incomplete (potentially split) UTF-8 byte sequences *at the end of* any given string buffer.
- Returns 0, if no such incomplete UTF-8 sequence is encountered. Otherwise, it returns the position of the first byte of the incomplete sequence, relative to the end of the buffer. E.g., the function will return 2, if it encounters a three-byte sequence with a single byte missing.

What it doesn't:
- Traverses the whole buffer to check for invalid sequences (CheckUtf8 already does that).

The main purpose of this function is to allow developers to work better on environments where incoming data chunks are not always aligned  with the variable length of UTF-8 sequences.

Please review.


uppsrc/Core/Utf.cpp
```diff
@@ -7,6 +7,17 @@ bool CheckUtf8(const char *s, int len)
 	return FromUtf8_([](wchar) {}, s, len);
 }
 
+int CheckUtf8Split(const char *s, int len)
```
mirek-fidler, line 10: Do we really need to add such poorly defined (at the end? Somehow I cannot find real world usage) function just because of test?

ismail-yilmaz: > Do we really need to add such poorly defined (at the end? Somehow I cannot find real world usage) function just because of test?

I can understand  this, but I am not proposing this just because of a test.  Especially the text mode applications, using;

- ssh or other binary protocols to move data back and forward
- local and remote applications with heavy data IO.

suffers from this frequently. This is more pronounced in text mode applications with high throughput (such as modern terminal apps using ncurses or other text user interface, eg. btop, htop). Almost all modern terminals (including TerminalCtrl), multiplexers and some apps does this check in order to prevent or minimize possible data corruption. Not to mention web apps such as Chromium, Firefox et al have similar helper functions or methods.

ismail-yilmaz:
Closing this.

PR #105, opened 2022-10-02
Fix to compile on POSIX by adding missing clang fucntions

klugier:
Fix to compile on POSIX by adding missing clang fucntions. Moreover, all headers from libclang has been added to avoid unnecessary work related to modifying libclang.h all the time when something new is added.


mirek-fidler:
This shows clear miunderstanding of ide/libclang internals.

PR #113, opened 2022-10-31
Eliminating compilation warning from CtrlLib::PopUpList class

klugier:
friend declaration should do not mix class with struct. So, if the parent structure is declare as struct in friend operation we should use exactly the same keyword. In the context of our code PopUpList has this problem.

Tested with ARM Clang on macOS.


ped7g (approved):

mirek-fidler:
Fixed independently.

PR #118, opened 2022-12-11
one possible bugfix, and few more small changes to lower amount of warnings while building ide

ped7g:
build platform: KDE neon (based on Ubuntu 22.04), clang 14.0.0 in C++20 mode
(except changing build method to -std=c++20, it's the default CLANG build method after cloning git repository and using make to build IDE from command line)

There are still some warnings left, but some of those can't be resolved as cleanly, the `|` vs `||` would require extra casting to `int`, and some unused variables are used for diagnostic or debug builds. Also I avoided Assist.cpp and Navigator.cpp just in case Mirek is still editing them a lot.


mirek-fidler:
Thanks, applied (individually...).

ped7g:
did you push it to github? Looks to me like not, or I'm confused... (not important for me, just willing to rebuild IDE for further testing with that "advent of code" thing)

PR #119, opened 2022-12-13
possible fix

ped7g:
Seems you left out the most important change to fix the `sOperatorTab` access with signed char.

Was that intentional? To me it looks like logical fix, although I didn't verify how to trigger the bug by source code.

(the change in Mt.cpp is not important, but the Signature.cpp IMO is)


mirek-fidler:
Sorry, that was accident.

ped7g:
fixed on master by Mirek.

PR #124, opened 2022-12-25
remove bitwise or with booleans warnings by changing to plus

ped7g:
bool(bool(a) | bool(b)) == bool(bool(a) + bool(b)), but without warning suggesting casting to integer.

### DIFF
--- a/uppsrc/CtrlCore/GtkEvent.cpp
+++ b/uppsrc/CtrlCore/GtkEvent.cpp
@@ -638,7 +638,7 @@ void Ctrl::Proc()
 			if(x)
 				kv = x->b;
 			else {
-				if(GetCtrl() | GetAlt()) {// fix Ctrl+Shift+1 etc...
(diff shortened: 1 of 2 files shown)


uppsrc/CtrlCore/GtkEvent.cpp
```diff
@@ -638,7 +638,7 @@ void Ctrl::Proc()
 			if(x)
 				kv = x->b;
 			else {
-				if(GetCtrl() | GetAlt()) {// fix Ctrl+Shift+1 etc...
```
mirek-fidler, line 641: What about plain old ||?

ped7g: `GetCtrl()` and `GetAlt()` seems to have no side-effects, so short-cutting it with `||` should work well, and then it makes me wonder why it was not used already in original source.

Will change the PR.

mirek-fidler: Very likely just harmless typo.

PR #125, opened 2022-12-26
Doesn't work without casting to correct pointers

emakei:


mirek-fidler:
Bug is legit, thank you. Fix not quite so, fixed correctly.

PR #126, opened 2022-12-31
A Crude X11 fix with a chance to compile and test.

arilect:


uppsrc/CtrlCore/CtrlCore.h
```diff
@@ -685,6 +685,7 @@ class Ctrl : public Pte<Ctrl> {
 	void    SyncScroll();
 	void    Refresh0(const Rect& area);
 	void    PaintCaret(SystemDraw& w);
+	int16        caretx, carety, caretcx, caretcy;//8  //aris002
```
klugier, line 688: No need for introducing new variable there is:
`static Rect      caretRect;`

As I wrote please reference to this commit https://github.com/ultimatepp/ultimatepp/commit/c7c4d97499be4b3f4a6423103188cabe2ff3e570

uppsrc/CtrlCore/X11Ctrl.cpp
```diff
@@ -6,6 +6,7 @@ namespace Upp {
 
 void Ctrl::GuiPlatformConstruct()
 {
+	caretcx = caretcy = caretx = carety = 0; //aris002
```
klugier, line 9: Rect should be constructed automatically.

uppsrc/CtrlCore/X11Ctrl.cpp
```diff
@@ -92,12 +93,12 @@ void GuiPlatformAfterMenuPopUp()
 	Ctrl::ProcessEvents();
 }
 
-void Ctrl::PaintCaret(SystemDraw& w)
-{
-	GuiLock __;
-	if(this == caretCtrl && WndCaretVisible)
-		w.DrawRect(caretx, carety, caretcx, caretcy, InvertColor);
-}
+//void Ctrl::PaintCaret(SystemDraw& w)  //aris002
```
klugier, line 96: No need for annotation //aris002. You will be present in the git history, so this is usless and we will need to remove this comment anyway :) Please remove it in other places in this PR.

uppsrc/CtrlCore/X11Ctrl.cpp
```diff
@@ -92,12 +93,12 @@ void GuiPlatformAfterMenuPopUp()
...
+//{
+//	GuiLock __;
+//	if(this == caretCtrl && WndCaretVisible)
+//		w.DrawRect(caretx, carety, caretcx, caretcy, InvertColor);
+//}
```
klugier, line 101: Commented code is bad - if we do not need this functionality just remove the code completely. Please apply this comment to all places in this PR.

uppsrc/CtrlCore/X11Ctrl.h
```diff
@@ -94,7 +94,9 @@
 
 	virtual void    EventProc(XWindow& w, XEvent *event);
 	virtual bool    HookProc(XEvent *event);
-	Window  GetWindow() const         { return top ? top->window : None; }
+	Window  GetWindow() const         { return top && utop ? utop->window : None; }  //aris002
+	static int    GetCaretBlinkTime()               { return 500; } //aris002
+    void SetCaret(int x, int y, int cx, int cy); //aris002
```
klugier, line 99: SetCaret method should be completely removed it is no longer used in GTK and Win32 implementation.

uppsrc/CtrlCore/X11Ctrl.h
```diff
@@ -94,7 +94,9 @@
 
 	virtual void    EventProc(XWindow& w, XEvent *event);
 	virtual bool    HookProc(XEvent *event);
-	Window  GetWindow() const         { return top ? top->window : None; }
+	Window  GetWindow() const         { return top && utop ? utop->window : None; }  //aris002
```
klugier, line 97: IMO, no need for checking top and utop:
`Window GetWindow() const { return top ? utop->window : None; }`
should be sufficient. Please look at GetTop() and SetTop() implementation.

uppsrc/CtrlCore/X11DHCtrl.cpp
```diff
@@ -274,28 +274,29 @@ void DHCtrl::Terminate(void)
 
 	// Unmaps the window
 	MapWindow(false);
-
+	if(utop) {
```
klugier, line 277: GetTop() is more secure than checking utop only:
`Top         *GetTop()               { return top ? utop : NULL; }`
You can even probably use top here only....

Consider converting to early return
```
if (!top) {
    // Not sure what we should set here in such sitatuion....
    return;
}
```

uppsrc/CtrlCore/X11DHCtrl.cpp
```diff
@@ -274,28 +274,29 @@ void DHCtrl::Terminate(void)
...
+		// Removes XWindow from Upp list
+		RemoveXWindow(utop->window);
+	
+		// Destroys the window
+		// Not to do, it's done destroying the parent window by X11 system
+	//	XDestroyWindow(Xdisplay, top->window);
+	
+		// Destroys created Top struct
+		delete utop;
+		utop = NULL;
```
klugier, line 297: Use build in DeleteTop() instead. It is also setting top flag, which is missing in this implementaiton.

uppsrc/CtrlCore/X11Proc.cpp
```diff
@@ -418,17 +420,20 @@ void Ctrl::EventProc(XWindow& w, XEvent *event)
...
-		Xeventtime = event->xmotion.time;
-		Point p = mousePos - Xbuttonpos;
-		if(max(abs(p.x), abs(p.y)) > 4)
-			Xbuttontime = Xeventtime - 0x80000000;
-		sModState = event->xmotion.state;
-		DispatchMouse(MOUSEMOVE, Point(event->xmotion.x, event->xmotion.y));
-		DoCursorShape();
-		break;
+		Top *top = GetTop(); //aris002
+		if(top) {
```
klugier, line 424: Remove this line:
`Top *top = GetTop(); //aris002`
and replace with
```
if (!top) { // top is class variable responsible for checking if utop is set....
    break;
}

/// refer to utop variable instead of top...
```

uppsrc/CtrlCore/X11Top.cpp
```diff
@@ -321,7 +321,8 @@ void TopWindow::Open(Ctrl *owner)
 	XSetClassHint(Xdisplay, GetWindow(), class_hint);
 	LLOG("WndShow(" << visible << ")");
 	WndShow(visible);
-	if(visible) {
+	Top *top = GetTop(); //aris002
+	if(top && visible) { //aris002
```
klugier, line 325: The same use top and utop variables from Ctrl class instead... TopWindow is friend of Ctrl....

uppsrc/CtrlCore/X11Wnd.cpp
```diff
@@ -55,7 +55,8 @@ static int s_starttime;
 void Ctrl::DoPaint(const Vector<Rect>& invalid)
 {
 	GuiLock __;
-	if(IsVisible()) {
+	Top *top = GetTop();  //aris002
+	if(top && IsVisible()) {
```
klugier, line 59: The same do not use top. Use top and utop variables from Ctrl class.

uppsrc/CtrlCore/X11Wnd.cpp
```diff
@@ -109,6 +110,7 @@ Ctrl *Ctrl::CtrlFromWindow(Window w)
 Ctrl::XWindow *Ctrl::GetXWindow()
 {
 	GuiLock __;
+	Top *top = GetTop(); //aris002
```
klugier, line 113: Same...

uppsrc/CtrlCore/X11Wnd.cpp
```diff
@@ -121,32 +123,34 @@ Window Ctrl::GetParentWindow(void) const
...
 // Get control with parentwindow as handle
-Ctrl *Ctrl::GetParentWindowCtrl(void) const
+Ctrl *Ctrl::GetParentWindowCtrl(void) const   //aris002 just blindly changing
 {
 	GuiLock __;
-	Ctrl *q = parent;
-	while(q && !q->top)
-		q = q->parent;
+	Ctrl *q = uparent;
+	while(q && !q->utop)
```
klugier, line 140: You should chec top instead of utop here.... These problems are present till the end of the PR, I will do not add comments. Please fix...

uppsrc/CtrlCore/CtrlCore.h
arilect: 1. As we spoke on Discord and you said "The first commit on the list should be related to changes that broke pure X11 compilation." This exact line had been used in Ctrl structure before and now it exactly restores the compilation ability, gives a chance to debug and investigate possible harmonisation with other caret related changes.  2. Now I will try to trace those changes. 3. Also, this is exactly the 'strategic' problem I said I wanted to discuss with CtrlCore chief architects before committing this code.  4. On the other hand, that is why I proposed to create a separate X11 dev  branch on ultimatepp/ultimatepp for more developers to participate. And I wanted to make live easier/faster by creating a compile/debug-able branch. 5. But I am open to upp or your style of work. P.S. I checked again and this line and related things can be removed.

uppsrc/CtrlCore/X11Ctrl.cpp
arilect: Same as https://github.com/ultimatepp/ultimatepp/pull/126#discussion_r1059698714

uppsrc/CtrlCore/X11DHCtrl.cpp
arilect: Yes. I have DeleteTop() in my other changes. I just left it in the first block of changes and to trace the top to utop renaming and to sync with macos-cocoa renaming.

uppsrc/CtrlCore/X11DHCtrl.cpp
arilect: Probably, yes. But... I am trying to trace that bool top used in Ctrl.h and all those 
And I do not like
` bool Ctrl::IsWndOpen() const {
	GuiLock __;
//	DLOG("IsWndOpen " << Upp::Name(this) << ": " << (bool)top);
	return top;
} `
And one more thing that bool topmost also exists...
I would suggest maybe to decide different concepts and different naming. 
and maybe to rename istop or isopen or ismapped
On the other hand, it would be logical to use here IsWndOpen() - the same way you suggested to use DeleteTop() . And btw, it was just if(top) and I renamed into if(utop). So utop might exist but not mapped therefore just top==false?

uppsrc/CtrlCore/X11Proc.cpp
arilect: Same as https://github.com/ultimatepp/ultimatepp/pull/126#discussion_r1059700681 .  But  please check in how often is this line (Top *top = GetTop();)  used in other CtrlCore places...  But my main question still : X11 Window might exist, therefore utop==true but  top==false?  or what about topmost?

uppsrc/CtrlCore/X11Top.cpp
arilect: Same as https://github.com/ultimatepp/ultimatepp/pull/126#discussion_r1059701436

uppsrc/CtrlCore/X11Wnd.cpp
arilect: Same as https://github.com/ultimatepp/ultimatepp/pull/126#discussion_r1059701436

klugier:
@arilect can you close this PR in favor of #127.

PR #127, opened 2023-01-01
A less crude fix to get CtrlCore compiled with X11 flag

arilect:


uppsrc/CtrlCore/X11Proc.cpp
```diff
@@ -418,17 +419,20 @@ void Ctrl::EventProc(XWindow& w, XEvent *event)
...
-		mousePos = Point(event->xmotion.x_root, event->xmotion.y_root);
-		Xeventtime = event->xmotion.time;
-		Point p = mousePos - Xbuttonpos;
-		if(max(abs(p.x), abs(p.y)) > 4)
-			Xbuttontime = Xeventtime - 0x80000000;
-		sModState = event->xmotion.state;
-		DispatchMouse(MOUSEMOVE, Point(event->xmotion.x, event->xmotion.y));
-		DoCursorShape();
-		break;
+		Top *top = GetTop();
```
klugier, line 422: This line should be removed:
if (top) // Global Ctrl variable... {
    while(XCheckWindowEvent(Xdisplay, utop->window, PointerMotionMask, event));
    ....

uppsrc/CtrlCore/X11Proc.cpp
```diff
@@ -418,17 +419,20 @@ void Ctrl::EventProc(XWindow& w, XEvent *event)
...
+			EndIgnore();
+			mousePos = Point(event->xmotion.x_root, event->xmotion.y_root);
+			Xeventtime = event->xmotion.time;
+			Point p = mousePos - Xbuttonpos;
+			if(max(abs(p.x), abs(p.y)) > 4)
+				Xbuttontime = Xeventtime - 0x80000000;
+			sModState = event->xmotion.state;
+			DispatchMouse(MOUSEMOVE, Point(event->xmotion.x, event->xmotion.y));
+			DoCursorShape();
+			break;
```
klugier: break should be after if statement. In this case it doesn't matter, but when there will be additiona case after MotionNotify it might lead to unpredictable behavior.

However, as I suggested we should use early return pattern here instead:
```
if (top) {
    break;
}

while(XCheckWindowEvent(Xdisplay, top->window, PointerMotionMask, event));
....
break;
```

uppsrc/CtrlCore/X11Ctrl.h
```diff
@@ -11,8 +11,8 @@
 
 private:
 	static ArrayMap<Window, XWindow>& Xwindow();
-	static int       WndCaretTime;
-	static bool      WndCaretVisible;
+//	static int       WndCaretTime;
+//	static bool      WndCaretVisible;
```
klugier: Please remove commented code. We do not need it and there is no change that after applying this PR it will be needed in the future.

uppsrc/CtrlCore/X11Ctrl.h
```diff
@@ -26,7 +26,7 @@
 	static void     TimerAndPaint();
 	static void     ProcessEvent(XEvent& event);
 	       void     Invalidate(XWindow& xw, const Rect& r);
-	static void     AnimateCaret();
+//	static void     AnimateCaret();
```
klugier: Same

uppsrc/CtrlCore/X11DHCtrl.cpp
```diff
@@ -197,8 +197,8 @@ bool DHCtrl::Init()
 	    			)
 	    : NULL;
 
-	top = new Top;
-	top->window = hwnd;
+	utop = new Top;
+	utop->window = hwnd;
```
klugier: You should use SetTop() here
```
Top* top = newTop;
top->window = hwnd;
SetTop(top);
```
SetTop also setting top to true, which is missing in this implementation.

GtkCreate.cpp - line 19 for reference.

uppsrc/CtrlCore/X11DHCtrl.cpp
```diff
@@ -274,28 +274,28 @@ void DHCtrl::Terminate(void)
...
-
-	// Frees input context as needed
-	if(cw->xic)
-	{
-		XDestroyIC(cw->xic);
-		cw->xic = NULL;
+		XWindow *cw = XWindowFromWindow(utop->window);
+	
+		// Frees input context as needed
+		if(cw->xic)
```
klugier: Shouldn't it be `if(cw && cw->xic)` { looks like risky situation for me...

uppsrc/CtrlCore/X11Wnd.cpp
```diff
@@ -55,14 +53,15 @@ static int s_starttime;
 void Ctrl::DoPaint(const Vector<Rect>& invalid)
 {
 	GuiLock __;
-	if(IsVisible()) {
+	Window xwin = GetWindow();
+	if(utop && IsVisible()) {
```
klugier, line 57: You should base on top variable not utop
```
if(top && IsVisible()) {
    // ...
```

uppsrc/CtrlCore/X11DHCtrl.cpp
arilect: Good point about the setting top to true.

uppsrc/CtrlCore/X11Proc.cpp
arilect: Yes.

uppsrc/CtrlCore/X11Ctrl.h
arilect: Done.

uppsrc/CtrlCore/X11Ctrl.h
arilect: Done.

uppsrc/CtrlCore/X11DHCtrl.cpp
arilect: Done.

uppsrc/CtrlCore/X11DHCtrl.cpp
arilect: Done.

PR #128, opened 2023-01-02
2023 kickstarter fix to compile CtrlCore and run upp apps with X11 flag

arilect:
All changes in one place now. You close the previous PR.


mirek-fidler:
Note: had to fix 'uparent' use. Not a good idea to use just one member of union without testing....

PR #129, opened 2023-01-02
Joint 2023-1 working fix for macos cocoa and X11. Mostly top to utop changes. Plus *Caret* things.

arilect:
Please delete other PR if necessary

### DIFF
--- a/uppsrc/Core/Core.h
+++ b/uppsrc/Core/Core.h
@@ -228,6 +228,9 @@
 #include <netdb.h>
 #include <netinet/in.h>
 //#include <libiberty.h>
+#ifdef __APPLE__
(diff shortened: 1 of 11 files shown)


klugier:
@arilect could we have a separate PR for mac only changes. Let's do not mess X11 changes with macOS cocoa changes. Mac version will be fixed by Mirek, so if we combine this two PR the changes that it will be merged to master are decreasing.

klugier (changes requested):
Please create separate PR for macOS changes.

arilect:
Have you missed that https://github.com/ultimatepp/ultimatepp/pull/128? That is exactly what are you requesting.

PR #138, opened 2023-01-22
Add simple RSAPublic/PrivateKey with method to handle RS256 signatures

lindquist:
This PR adds a simple module for handling RSA private/public keys, and for now a single method to sign/verify widely used RS256 signatures.


lindquist:
I generated the keys with this node script:

<pre>
const { generateKeyPairSync } = require("crypto");
const { privateKey, publicKey } = generateKeyPairSync("rsa", {
    modulusLength: 4096,
    publicKeyEncoding: {
        type: "pkcs1",
        format: "pem",
    },
    privateKeyEncoding: {
        type: "pkcs8",
        format: "pem",
    },
});
</pre>

mirek-fidler:
This looks very good. Do you think it would be possible to add reference example too? (I guess it will be almost the same as autotest, but anway).

lindquist:
Thanks. I've added a Reference example, I also went ahead and added support for generating keys, also writing PEM strings has been added, so it's much more complete now I think.
Adding support for other digests would be nice, but U++ does not have the hash functions, alternatively it should just use openssl to create the digest.
Also password encryption of keys is not supported.

lindquist:
The more I look at this, it seems 1) right move would be to move U++ to openssl3 2) openssl is a pain 3) The RSA functions I used are not only deprecated in openssl3 (like most everything), but also only support the older non-"SubjectPublicKeyInfo" format.
As a minimum, I think the module should be updated to use the EVP_PKEY APIs.

mirek-fidler:
Not glad to hear that. It is nightmare already to support 1 and 2... which
we ultimately need to.

lindquist:
I have created a repo "uppcloud" where I'm going to post these things. I'm going to take a look at supporting both RSA* and EVP_PKEY* keys there.
Then, perhaps we can look at if (some of) it should be merged into uppmain

lindquist:
> I have created a repo "uppcloud" where I'm going to post these things. I'm going to take a look at supporting both RSA* and EVP_PKEY* keys there. Then, perhaps we can look at if (some of) it should be merged into uppmain

... and the link: https://github.com/lindquist/uppcloud

lindquist:
I have added the beginning of an EVP based package:
https://github.com/lindquist/uppcloud/commit/5b7c7e37c2748fabff3b02589e9e34a9a8a49520

PR #140, opened 2023-01-23
AutoTest does not build on windows due to missing 'setenv'

lindquist:
This PR adds a new SetEnv function to match the GetEnv which can then be used in AutoTest
Test was a bit fiddly, feedback appreciated.

### DIFF
--- a/uppsrc/Core/App.h
+++ b/uppsrc/Core/App.h
@@ -1,4 +1,5 @@
 String  GetEnv(const char *id);
+bool SetEnv(const char *name, const char *value);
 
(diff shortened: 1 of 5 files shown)


uppsrc/Core/App.h
```diff
@@ -1,4 +1,5 @@
 String  GetEnv(const char *id);
+bool SetEnv(const char *name, const char *value);
```
mirek-fidler, line 2: I will add 3 spaces between bool and SetEnv... :)

PR #143, opened 2023-02-01
Editor overwrite mode fixes

ped7g:
Overwrite worked only in CSyntax files, failed in txt, etc..

second commit: seems to me there was ignored `count` argument in CSyntax::IndentInsert0 - I'm not sure how test this and where `count > 1` can emerge, but seems to me like bug.

Please review, I don't understand all possible code paths how these can be triggered, and why the `canoverwrite` argument even was added in the first place (as InsertChar itself does ignore overwrite for certain characters like newline), so I tried my best guess where the canoverwrite change should be harmless, but I have no deep understanding.


mirek-fidler:
Thanks.

PR #147, opened 2023-02-11
CoWork warnings with clang/C++17 (std::result_of)

lindquist:
std::result_of is deprecated in c++17 not 20


mirek-fidler:
Thanks!

PR #148, opened 2023-02-17
Minor SelectPkg(TheIde) polishing

M0untacir:


uppsrc/ide/SelectPkg.cpp
```diff
@@ -178,6 +178,7 @@ SelectPackageDlg::SelectPackageDlg(const char *title, bool selectvars_, bool mai
 	cancel.WhenAction = WhenClose = THISBACK(OnCancel);
 	clist.Columns(4);
 	clist.WhenEnterItem = clist.WhenKillCursor = THISBACK(ListCursor);
+	alist.AutoHideSb();
```
klugier, line 181: I was thinking about it, however I have never decided to change that.

uppsrc/ide/SelectPkg.cpp
```diff
@@ -212,7 +213,6 @@ SelectPackageDlg::SelectPackageDlg(const char *title, bool selectvars_, bool mai
 	brief <<= THISBACK(SyncBrief);
 	search.NullText("Search (Ctrl+K)", StdFont().Italic(), SColorDisabled());
 	search << [=] { SyncList(Null); };
-	search.SetFilter(CharFilterDefaultToUpperAscii);
```
klugier, line 215: I think this filter with making everything capital latter is fine and shouldn't be changed...

uppsrc/ide/ide.lay
```diff
@@ -1,7 +1,7 @@
 LAYOUT(SelectPackageLayout, 816, 468)
 	ITEM(Upp::ParentCtrl, parent, HSizePosZ(4, 8).VSizePosZ(4, 32))
-	ITEM(Upp::DropList, kind, Tip(t_("Select which packages should be displayed")).LeftPosZ(4, 124).BottomPosZ(7, 19))
-	ITEM(Upp::DropList, nest, Tip(t_("Select which packages should be displayed")).LeftPosZ(132, 112).BottomPosZ(7, 19))
+	ITEM(Upp::DropList, kind, Tip(t_("Select which packages should be displayed")).LeftPosZ(4, 124).BottomPosZ(6, 20))
+	ITEM(Upp::DropList, nest, Tip(t_("Select which packages should be displayed")).LeftPosZ(132, 112).BottomPosZ(6, 20))
```
klugier, line 4: What is the issue here? Can you explain or optimally put screenshot before/after.

klugier:
Hello, please add description to the PR and rational what you have changed and why. Without it it will be difficult to accept this PR and some changes are more about personal taste.

klugier (changes requested):
Please provide your motivation about changes introduced in this PR.

M0untacir:
Hi,
I just started diving into TheIde's code (to learn) and i thought i could make this dialog visually polished that's all.

- The auto hide makes that list less crowded.
- The uppercase filter was tripping when i first opened TheIde, it took me second to realize it wasn't my keyboard. It's not expected.
- As for the DropLists height, i made them to match the buttons height.

Sorry for not including a description, this is probably my second PR ever.
Thanks.

uppsrc/ide/SelectPkg.cpp
klugier: I remembered why I decided to do not propose this change. It is causing flickering when you switch from assembly with relatively small amount of packages (scrollbar not needed) to the assembly that has a lot (scrollbar needed).

M0untacir: I don't have flickering, i use Windows 10.

klugier: This is exactly what I mean - this is visually inconsistency. It looks strange when one time you have scrollbar and the other time you don't. In the end the placement of ArrayCtrl lines are different between these two views, which is not good and disrupting. So, for now I am against this change.

We need different solution to make it better. It is called overlay scrollbars. It is not yet implemented in CtrlLib, but it should give us space in many controls and one of the beneficent would be "Select main package" window.

---
Please, don't take my comments personal. I am glad that you created the PR and give the space for discussion. However, we can not accept everything.

M0untacir: Yes, overlay scrollbars would be the solution.
No worries, I understand.
Thank you.

PR #149, opened 2023-02-18
Update scripts.cpp

9t8:
tested individual commands on mageia 8 vm
urpmi should not have `install` passed to it
`clang-tools-extra` is necessary to install `clang-format`


9t8:
maybe urpmi support should be removed. the VM I used also came with dnf (the Wikipedia page of mageia lists urpmi as legacy), as would I imagine most others, and in these cases the install script would select dnf anyway

9t8:
more evidence supporting urpmi's removal is that no one seems to use it: I can't find anyone complaining about the fact that the command has been broken for years

PR #168, opened 2023-10-20
Add GetWidth() to Bar (analogous to GetHeight())

schutm:
I wasn't sure enough to also include a GetSize() as I don't know what consequences this would have. A GetPaneSize() would be another option. But since I didn't require the height I didn't pursue this option.


mirek-fidler:
Sounds good, except maybe both GetWidth and GetHeight should in fact be in ToolBar? (They seems to actually return ToolBar size).

ortman:
The GetWidth() method was added to master branch 2 years ago.
Let's close this pull request.

PR #170, opened 2023-10-20
core.upt should have an option for ssl

klugier:
SSL is requiered for all modern network application. For example today most of the http resources are available through https not http. When creating core package there should be an easy way to add SSL support to it


mirek-fidler:
OK, why not. That said, I am starting to think that maybe something like check-list of all or some available packages could be present there... Maybe in the future, this is good as it is.

PR #173, opened 2023-11-26
Ide: Git - GetCredentials fix (now returns true on success)

ismail-yilmaz:
Ide::GetCretentials returns only false, which causes git commits and pushes to fail.
This pull request aims to fix the issue.


mirek-fidler:
I had to deprecate credentials in Sync, probably will need to remove them completely. The problem is that they do not work with branches very well.

PR #177, opened 2023-11-29
IDE changes for Flatpak sandbox

klugier:
This PR's adapts TheIDE to be run under Flatpak. Please review, for me for now bot version works flatpaked and normal. However, very careful review will be extremely valuable.

flagSandboxFlatpak flag is passed via uppsrc/uppconfig.h during flatpak build after cloning the repository.

Due to high amount of commits, please merge using squash option.


uppsrc/ide/Debuggers/Gdb.cpp
```diff
@@ -599,9 +599,10 @@ bool Gdb::Create(Host& host, const String& exefile, const String& cmdline, bool
 
 #ifdef PLATFORM_POSIX
 #ifndef PLATFORM_MACOS
-	IGNORE_RESULT(system("setxkbmap -option grab:break_actions")); // to be able to recover capture in breakpoint
+	IGNORE_RESULT(HostSys("setxkbmap -option grab:break_actions")); // to be able to recover capture in breakpoint
 	String xdotool_chk = ConfigFile("xdotool_chk");
-	if(!FileExists(xdotool_chk) && system("xdotool key XF86Ungrab")) {
+	String out;
```
mirek-fidler, line 604: Does not seem to be used.

uppsrc/ide/clang/clang.cpp
```diff
@@ -154,8 +157,9 @@ bool Clang::Parse(const String& filename_, const String& content,
 	                                options & PARSE_FILE ? 0 : (filename2.GetCount() ? 2 : 1),
 	                                options);
 
-	if(!tu)
+	if(!tu) {
```
mirek-fidler: Really? :)

uppsrc/Core/config.h
```diff
@@ -147,6 +147,10 @@
 	#endif
 #endif
 
+#ifdef flagSandboxFlatpak
+	#define SANDBOX_FLATPAK 1
```
mirek-fidler: Maybe we can do runtime check instead? I can imagine IsRunningInFlatpak function in Core, then the binary would be "universal".

uppsrc/ide/Core/Host.h
```diff
@@ -2,9 +2,19 @@ enum { REMOTE_TIMEOUT = 2000 };
 
 extern String HostConsole;
 
+String WrappCmdlineForSandbox(const char* cmdline);
```
mirek-fidler: Is this used anywhere?

uppsrc/ide/Core/Host.h
```diff
@@ -2,9 +2,19 @@ enum { REMOTE_TIMEOUT = 2000 };
 
 extern String HostConsole;
 
+String WrappCmdlineForSandbox(const char* cmdline);
 void AddHostFlags(Index<String>& cfg);
 
 struct Host {
+	class SandboxUtils {
```
mirek-fidler: I hate adding whole subclass to add single single function....

uppsrc/ide/clang/clang.cpp
klugier: Will be removed!

uppsrc/ide/Debuggers/Gdb.cpp
klugier: HostSys requires second parameter as String reference to return int not bool. The HostSys with two parameters(cmd, out) is the equivalent for system in the context of return value.

uppsrc/ide/Core/Host.h
klugier: Yes, you are right. It should be removed.

uppsrc/ide/Builders/Install.cpp
```diff
@@ -172,7 +172,7 @@ void CreateBuildMethods()
 		SaveFile(bm_path, bm);
 	}
 #else
-	bool openbsd = ToLower(Sys("uname")).Find("openbsd") >= 0;
+	bool openbsd = ToLower(HostSys("uname")).Find("openbsd") >= 0;
```
klugier: Bug: HostSys is not available in this place. Host::CMDLINE_PREFIX should be used. I will rework this.

uppsrc/ide/Core/Host.cpp
```diff
@@ -5,8 +5,19 @@
...
 #define METHOD_NAME "Host::" << UPP_FUNCTION_NAME << "(): "
 
-Host::Host()
+#ifdef FLATPAK
+const char* Host::CMDLINE_PREFIX = "host-spawn --no-pty ";
+#else
+const char* Host::CMDLINE_PREFIX = "";
+#endif
+
+String Host::AddCmdlinePrefix(const char* cmdline)
```
mirek-fidler: You seem to be using CMDLINE_PREFIX for the same role as AddCmdlinePrefix, except it does String conversion. If you CMDLINE_PREFIX was String instead of const char *, I think this function could be removed. Less clutter, better code.

PR #178, opened 2023-12-02
DLOGs and DDUMPs are renamed in X11Wnd.cpp for successfull NOGTK builds.

ismail-yilmaz:
X11 (NOGTK) release builds fail due to the remaining DDUMPs and DLOGs in the code. They are renamed as LDUMP and LLOG macros.


ismail-yilmaz:
Is there any reason not to merge this? Nightly builds are still failing to compile with NOGTK flag.

klugier:
@ismail-yilmaz why do not remove this logs completely? I do not think they are needed and they will pollute log output. In general I agree that the compilation should be fixed even if pure X11 backend is not a priority for now.

FYI, @mirek-fidler.

mirek-fidler:
Reason to convert to L variant are reasonable - we eventually might need then to debug the code in future.

PR #189, opened 2024-02-11
CtrlCore: Added Horizontal mouse wheel wheel support via K_SHIFT.

User0755:
CtrlCore: Added Horizontal mouse wheel wheel support via K_SHIFT. This is available for both Win32 and POSIX.


uppsrc/CtrlCore/Win32Proc.cpp
```diff
@@ -340,14 +341,17 @@ LRESULT Ctrl::WindowProc(UINT message, WPARAM wParam, LPARAM lParam) {
 		}
 		return 0L;
 	case 0x20a: // WM_MOUSEWHEEL:
+	case 0x020E: //#define WM_MOUSEHWHEEL                  0x020E
```
klugier: Why we are using hex values here? Can we use defined values?

User0755:
Hi, I guess some older Win SDK did not have support for WM_MOUSEWHEEL, so hex is and was used for widest compatibility.

Tom

Sent from Proton Mail mobile

-------- Original Message --------

klugier:
In accordance to documentation WM_MOUSEWHEEL was introduced in WIndows Vista (https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousehwheel). Since we officially do not support Windows 7 there is no need to keep hex values for compatibility reasons.

The same is true for 0x20a. It was introduced in Windows 2000. SO, the same situation we should replaced it with defined. This code might remembered the times pre WIndows 2000 :) Source: https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousewheel.

User0755:
I was merely following the current path here... Now I tried to switch to using WM_MOUSEHWHEEL, but it seems that default CLANG tools on Windows do not support it.

Anyway, if Mirek chooses to change the strategy some way, he will do it himself in seconds.

mirek-fidler:
I do not quite like K_SHIFT solution. I think that what we should do is to add

	virtual void   MouseWheel(Point p, Point delta, dword keyflags);

with default implementation calling existing vertical MouseWheel. If the same name is problem (to be tested, I think it should be fine), name it MouseWheels.

Also missing win32.

mirek-fidler:
> In accordance to documentation WM_MOUSEWHEEL was introduced in WIndows Vista (https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousehwheel). Since we officially do not support Windows 7 there is no need to keep hex values for compatibility reasons.

Hex values with comment are not that much problem, possible incompatibility with build tools is.

User0755:
Win32 is only missing, because it was conflicted by the merged pen fix in Win32Proc.cpp.

Changing mouse wheel data from "int zd" to "Point delta" will be a bit larger task. It involves everything from DoMouse up to MouseWheel.

If you prefer this approach, I can take a look if I can do it... Not sure yet.

//Tom

EDIT: At least the following would require changes, probably more:

DoMouse()
DispatchMouse()
DispatchMouseEvent()
MEvent0()
MouseEventH()
FrameMouseEventH()
mousehook
ChildFrameMouseEvent()
MouseEvent0()
MouseEvent()
ChildMouseEvent()
LogMouseEvent()

Is this the way to go? I'm afraid this might break a lot of things.

User0755:
Hi, it seems I have accidentally closed this PR. Well, I guess, I'll just have to open a new one after creating a new branch for it. (Not so many hours on github yet, so all kinds of weird stuff happens to me here...)
Anyway, Mirek, what do you think about the vast number of affected locations in the code if the mouse wheel reading in 'int zd' is switched to 'Point delta'? (Please look at my previous edited posting.) I'm a bit worried about the consequences.
// Tom

PR #190, opened 2024-02-11
CtrlCore: Fix pen detection in Win32

User0755:
CtrlCore: Fix pen detection in Win32 properly passing through touch screen gestures as mouse emulation.

### DIFF
--- a/uppsrc/CtrlCore/Win32Proc.cpp
+++ b/uppsrc/CtrlCore/Win32Proc.cpp
@@ -91,7 +91,7 @@ LRESULT Ctrl::WindowProc(UINT message, WPARAM wParam, LPARAM lParam) {
 
 	cancel_preedit = DoCancelPreedit; // We really need this just once, but whatever..
 
-	is_pen_event = (GetMessageExtraInfo() & 0xFFFFFF00) == 0xFF515700;
+	is_pen_event = (GetMessageExtraInfo() & 0xFFFFFF80) == 0xFF515700; // https://learn.microsoft.com/en-us/windows/win32/tablet/system-events-and-mouse-messages?redirectedfrom=MSDN
 	
 	POINT p;
 	if(::GetCursorPos(&p))


uppsrc/CtrlCore/Win32Proc.cpp
```diff
@@ -91,7 +91,7 @@ LRESULT Ctrl::WindowProc(UINT message, WPARAM wParam, LPARAM lParam) {
 
 	cancel_preedit = DoCancelPreedit; // We really need this just once, but whatever..
 
-	is_pen_event = (GetMessageExtraInfo() & 0xFFFFFF00) == 0xFF515700;
+	is_pen_event = (GetMessageExtraInfo() & 0xFFFFFF80) == 0xFF515700; // https://learn.microsoft.com/en-us/windows/win32/tablet/system-events-and-mouse-messages?redirectedfrom=MSDN
```
klugier, line 94: I think the pen detection is done properly. In accordance to documentation:
`The value returned from [GetMessageExtraInfo](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getmessageextrainfo) needs to be mask-checked against 0xFFFFFF00, and then compared with 0xFF515700. The following definitions may make this clearer:`

I think we need to have separate variable like is_touch_event in addition to pen:
`Vista, the eighth bit, masked by 0x80, is used to differentiate touch input from pen input (0 = pen, 1 = touch).`

So, in this place we should:
```
auto msg_info = GetMessageExtraInfo();
is_pen_event = (GetMessageExtraInfo() & 0xFFFFFF00) == 0xFF515700;
is_touch_event = (GetMessageExtraInfo() & 0xFFFFFF00) == 0xFF515700;
```

User0755: No. The thing is that the 'standard' pen detection mask actually catches both pen and touch as pen. As a result, the touch gestures get pen processing and mouse events emulated from touch later get trapped as they are expected to be handled by pen processing, which they are not (and should not).

As you noticed above too, the 0x80 mask since Vista will have 1 for touch and 0 for pen. Therefore, the true pen detection becomes:

is_pen_event = (GetMessageExtraInfo() & 0xFFFFFF80) == 0xFF515700;

,as I have suggested in my PR. If touch detection is required, it should be

is_touch_event = (GetMessageExtraInfo() & 0xFFFFFF80) == 0xFF515780;

However, we do not need this for anything right now.

Also, please note that I have been testing this for about a month now with touch screen, pen screen and an external Wacom pen tablet. And this works with pen enabled U++ applications.

PR #192, opened 2024-02-13
CtrlCore: Horizontal mouse wheel support using MouseWheelHor()

User0755:
This solution (branch mousewheel3) uses a separate mouse event (MOUSEHWHEEL) for propagating horizontal mouse wheel events. This is quite straight forward and avoids major changes in CtrlCore. Also, this will not cause compatibility issues for existing code. The Ctrls will receive the event as a call to MouseWheelHor(). K_SHIFT is not manipulated.

--

I have also created another branch (mousewheel2) that attempts to pass the mouse wheel in a Point structure. However, this path proved to behave like a snowball rolling down the hill causing a lot of changes all around. In the end, I was not able to make the call to MouseWheel() with Point delta while still supporting current MouseWheel() with int. The reason was that the default Ctrl::MouseWheel() had to exist for propagating unprocessed calls for other Ctrls. So, MouseWheel() with Point delta is not compatible with existing code base. It is worth noting that both Windows and GTK produce the vertical and horizontal wheel events separately, so passing them combined as a point would not yield much benefit. (This would only happen after event compression.)

Finally, the simplest solution of all (now available as branch mousewheel-K_SHIFT), is also available for evaluation on User0755/ultimatepp.

// Tom

### DIFF
--- a/uppsrc/CtrlCore/CtrlCore.h
+++ b/uppsrc/CtrlCore/CtrlCore.h
@@ -882,6 +882,7 @@ public:
 		MOUSELEAVE    = 0x30,
 		CURSORIMAGE   = 0x40,
 		MOUSEWHEEL    = 0x50,
+		MOUSEHWHEEL   = 0x60,
 
 		DOWN          = 0x80,
 		UP            = 0x90,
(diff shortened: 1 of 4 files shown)


mirek-fidler:
I do not much care about "...Hor" at the end of virtual method name so I reserve the right to rename it in the near future (MouseHWheel or MouseWheelH)....

User0755:
> I do not much care about "...Hor" at the end of virtual method name so I reserve the right to rename it in the near future (MouseHWheel or MouseWheelH)....

Absolutely no problem, go ahead. However, in that case I suggest selecting MouseWheelH since it is easier to find all MouseWheel* when searching across file/files.

Thanks for merging!

// Tom

PR #194, opened 2024-02-28
plugin/zip: Zip64 support for UnZip and fixes for Zip

User0755:
Hi,

Here's Zip64 support for UnZip. Unfortunately, this is more difficult to test than Zip. At this time I have only tested with Zip files created with plugin/zip. If you have other ways to test Unzip, please do.

// Tom

### DIFF
--- a/uppsrc/plugin/zip/UnZip.cpp
+++ b/uppsrc/plugin/zip/UnZip.cpp
@@ -9,13 +9,15 @@ void UnZip::ReadDir()
 	file.Clear();
 	current = 0;
 
-	int entries = -1;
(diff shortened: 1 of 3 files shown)


User0755:
The last commit -- while really applying to Zip rather than UnZip -- improves zipping by properly handling large archive creation even with small files and folders when 4GB archive size has been exceeded.

BTW: String is an extremely handy container for dealing with large data such as file contents here in Zip and elsewhere, but unfortunately it gets exhausted at 2GB. Is a 64 bit String out of question...?  :)

// Tom

PR #195, opened 2024-04-07
Core: Added all well known http status codes

klugier:
I decided to add all well knowned http status code to Core. The main reason for that is to do not use magic numbers and magic strings and having everything under well defined constants. 

Status codes are basing on ints, so they are a good fit for the current code base.

This change also introduces HttpStatusLine for simplifying HttpResponse calls with status codes. In the future HttpResponse function will need to be replaced. One of the reason to do it is that it will be almost impossible to implement HTTP 2.0 protocol (Break api changes required).

The implementation is similar to http implementation in GoLang (https://go.dev/src/net/http/status.go).


mirek-fidler:
Not really sure whether this is useful or obsfuscation, but I guess a list of http status codes does not really hurt, even if in most cases they are not used in its original meanings anyway - I mean, we seldom do 'real' website https, but we do a lot of API http, rest-like or otherwise, and there these codes are pretty arbitrary and almost always not used in original meaning, phrases have to be different as well to better explain the problem (and that is true even if U++ code is just using 3rd party API).

What I think is pretty ugly overengineering is HttpStatusLine class. I would just add another overload of HttpResponse without phrase parameter and supply phrase by HttpStatus::ToString. Much less redundant...

mirek-fidler:
Another comment: This should be reimplemented using HttpStatusCode.i file with

CODE(OK, "OK")

and then include this twice to get enum and ToString. Less source code and much easier to maintain.

klugier:
Hello Mirek,

Thanks for review! I did as you suggested.

Regards,
Klugier

PR #207, opened 2024-09-29
reference/SshBasics: Ssh polymorphism/RTTI example for the SshBasics reference example set.

ismail-yilmaz:
This patch aims to add a polymorphism & RTTI example to the SShBasics reference examples set.

Demonstrates:
- Creating three different channels (Scp, SFtp, and SshExec) in the same Ssh object array,
- Identifying and converting the stored Ssh objects to the proper channels respectively and calling them.


reference/SshBasics/main.cpp
```diff
@@ -58,8 +58,11 @@ CONSOLE_APP_MAIN
 		ForwardTcpIp(session);
 #elif defined(SSH_PICK_SEMANTICS)
 		SshPick(session);
+#elif defined(SSH_POLYMORPHISM)
```
klugier, line 61: Where is SSH_POLYMORPHISM flag defined?

ismail-yilmaz: > Where is SSH_POLYMORPHISM flag defined?

Users *can* define it if they would like to run the specific example. By default the `SSH_SFTP` flag is defined. (this has been the case since the beginning.)

klugier:
Do you want this change in 2024.1 or it should be deliver after?

ismail-yilmaz:
> Do you want this change in 2024.1 or it should be deliver after?

This can be merged immediately, since there is nothing that requires the latest version of libssh2. :)

PR #211, opened 2024-10-22
Ide: Fix compilation error and warnings related to MSVC 22

klugier:
TheIDE doesn't compile with the latest stable MSVC. This PR address this problem. Also, I decieded to change instant setup for MSVC to configure llvm dir by default. With that change MSVC should compile TheIDE out of the box, without the need for manual configuration.


mirek-fidler:
Thanks.

PR #212, opened 2024-10-29
reference: Added markdown example (with image support).

ismail-yilmaz:
As the title suggest, this pull request aims to add a simple markdown viewer example to reference examples section.

Rationale: 

a) Markdown package does not have any example code in the references section and b) even the existing code elsewhere (e.g., in TheIDE) does not demonstrate the following features of `plugin/md` package:

- Basic way of handling and displaying (local) images.
- Copying the resulting `qtf` data to clipboard so that it can be used elsewhere.

A screenshot of the example, displaying Bobcat's [README](https://github.com/ismail-yilmaz/Bobcat) file.

resim

Please review.


reference/MarkdownViewer/MarkdownViewer.upp
```diff
@@ -0,0 +1,15 @@
...
+	Painter,
+	plugin/jpg,
+	plugin/md;
+
+file
+	main.cpp;
+
+mainconfig
+	"" = "GUI",
+	"" = "GUI .NOGTK";
```
klugier: Not sure if we need this flag. One GUI flag should be enough here.

reference/MarkdownViewer/main.cpp
```diff
@@ -0,0 +1,73 @@
+#include <CtrlLib/CtrlLib.h>
+
+#include <plugin/md/Markdown.h>
+#include <plugin/bmp/bmp.h>
+#include <plugin/jpg/jpg.h>
+
+using namespace Upp;
+
+struct MarkdownViewer : TopWindow {
+	typedef MarkdownViewer CLASSNAME;
```
klugier: This is not needed. Just use lambda callback instead. It was deprecated some time ago...

reference/MarkdownViewer/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+	void ParseFile(const String& fname)
+	{
+		images.Clear();
+		MarkdownConverter mdc;
+		mdc.WhenImages = [&](VectorMap<String, String>& imglist)
+		{
+			String dir = GetFileDirectory(fname);
+			Progress pi(t_("Loading images..."), imglist.GetCount());
+			pi.pi.Percent();
+			for(int i = 0; i < imglist.GetCount(); i++) {
```
klugier, line 52: For each loop can be used instead.

reference/MarkdownViewer/MarkdownViewer.upp
```diff
@@ -0,0 +1,15 @@
+description "A simple markdown viewer with image displaying capability.\377";
```
klugier, line 1: What about providing same .md file ready to be open by the reference app? For example in examples/SVGView we are providing .svg files.

Something with lorem ipsum md with image embeeding should be fine.

klugier:
Good idea! Markdown reference is definitly need, as markdown is very popular text file format.

reference/MarkdownViewer/MarkdownViewer.upp
ismail-yilmaz: Done.

reference/MarkdownViewer/main.cpp
ismail-yilmaz: Done.

reference/MarkdownViewer/main.cpp
ismail-yilmaz: Is this necessary? I really don't like to bring in a std::library function when it is not more useful than the traditional for loop. IMO, it doesn't really add anything here.

reference/MarkdownViewer/MarkdownViewer.upp
ismail-yilmaz: Done. Now the app opens a default example file, with image support. Please check.

PR #214, opened 2024-11-06
Updated to new version for website

XKFS:


klugier:
This is not need now. Also, some other places will need be updated as well. I will do this changes when finally we will have 2024.1.

XKFS:
No problem, Thanks man.

PR #215, opened 2024-11-15
Core: Add Linear Interpolation Function for Common Types (Numbers, Point, Size, Rect)

ismail-yilmaz:
Linear interpolation (Lerp) is a widely used method for smoothly transitioning between two values. It's commonly applied in fields such as UI development, animations, and geometry manipulation.

This pull request introduces a generic `Lerp` function, along with its specializations for the following types:
- Numeric types (`int`, `float`, `double`)
- `Point`
- `Size`
- `Rect`

Additionally, a simple reference example is included to demonstrate the usage of the `Lerp` function.

Please review .


mirek-fidler:
Delayed until after 2024 release...

mirek-fidler:
Added Lert(Color) too.

What do you think about https://en.cppreference.com/w/cpp/numeric/lerp ? There are some additional guarantees, maybe worth implementing these too? (We are now implementing "naive_lerp" from the example, maybe the precision issue is worth consideration).

mirek-fidler:
constexpr float lerp(float val1, float val2, float amount) {
      // Don't do (val2 - val1) * amount + val1 as it has worse precision.
      return val2 * amount + val1 * (1.0f - amount);
    }

klugier:
Also, please do not forget to add documentation for the new functions :)

ismail-yilmaz:
> Added Lert(Color) too.
> 
> What do you think about https://en.cppreference.com/w/cpp/numeric/lerp ? There are some additional guarantees, maybe worth implementing these too? (We are now implementing "naive_lerp" from the example, maybe the precision issue is worth consideration).

Thanks for the review. I am aware of the std implementation. I proposed this template because it is useful in many practical casesin U++. However, If it need that much precision, I usually resort to professional math libraries. :) Then again, there's no harm in adding some checks. How about this one:

```
template <typename T>
inline T Lerp(T a, T b, double t)
{
    static_assert(std::is_arithmetic_v<T>, "Upp::Lerp: Only arithmetic types are supported.");

    if constexpr(std::is_floating_point_v<T>) {
        if(IsNaN(t) || IsInf(t)) { // this possibly prevents compile time evaluation.
	         return a;
        }
    }

    // Clamp to {0, 1} to prevent extrapolation
    t = clamp(t, 0.0, 1.0);

    // Prevent potential numeric overflow/underflow
    if constexpr(std::is_floating_point_v<T>) {
        // Use more precise floating-point interpolation
        return T(a * (1.0 - t) + b * t);
    }
    
    // For integral types, use double to prevent overflow
	return T(a + t * ((double) b - (double) a));
}

```

mirek-fidler:
I think you are overcomplicating it.

- I guess nan/inf issues are not very important, in any case I am rather more OK with Lerp returning NaN or Inf than fixing something.

- Not sure why we should prevent extrapolation, that was the first thing I liked about it... (even added it to reference example).

- Not sure whether we need to handle integers differently, float formula seems to work just fine IMO. I have even tried

#include <Core/Core.h>

using namespace Upp;

template <typename T>
inline T Lerp1(T a, T b, double t) {
	return T(a + t * ((double) b - (double) a));
}

template <typename T>
inline T Lerp2(T a, T b, double t) {
    return T(a * (1.0 - t) + b * t);
}

CONSOLE_APP_MAIN
{
	DDUMP(Lerp1(1, 1000, 0.999999));
	DDUMP(Lerp1(1, 1000, 1));
	DDUMP(Lerp1(1, 1000, 0));
	DDUMP(Lerp1(1, 1000, -0.00001));
	DDUMP(Lerp1(1, 1000, -0.01));
	DLOG("================");
	DDUMP(Lerp2(1, 1000, 0.999999));
	DDUMP(Lerp2(1, 1000, 1));
	DDUMP(Lerp2(1, 1000, 0));
	DDUMP(Lerp2(1, 1000, -0.00001));
	DDUMP(Lerp2(1, 1000, -0.01));
}

And of course I might be missing something but it seems to be fine.

So I am basically suggesting to use "more precise" formula and be done with it...

ismail-yilmaz:
> I think you are overcomplicating it.
> 
>     * I guess nan/inf issues are not very important, in any case I am rather more OK with Lerp returning NaN or Inf than fixing something.
> 
>     * Not sure why we should prevent extrapolation, that was the first thing I liked about it... (even added it to reference example).
> 
>     * Not sure whether we need to handle integers differently, float formula seems to work just fine IMO. I have even tried
> 
> 
> #include <Core/Core.h>
> 
> using namespace Upp;
> 
> template inline T Lerp1(T a, T b, double t) { return T(a + t * ((double) b - (double) a)); }
> 
> template inline T Lerp2(T a, T b, double t) { return T(a * (1.0 - t) + b * t); }
> 
> CONSOLE_APP_MAIN { DDUMP(Lerp1(1, 1000, 0.999999)); DDUMP(Lerp1(1, 1000, 1)); DDUMP(Lerp1(1, 1000, 0)); DDUMP(Lerp1(1, 1000, -0.00001)); DDUMP(Lerp1(1, 1000, -0.01)); DLOG("================"); DDUMP(Lerp2(1, 1000, 0.999999)); DDUMP(Lerp2(1, 1000, 1)); DDUMP(Lerp2(1, 1000, 0)); DDUMP(Lerp2(1, 1000, -0.00001)); DDUMP(Lerp2(1, 1000, -0.01)); }
> 
> And of course I might be missing something but it seems to be fine.
> 
> So I am basically suggesting to use "more precise" formula and be done with it...

Ah, OK, I misunderstood your question then. Yes, the alternative formula has better precision. Thanks!

PR #216, opened 2024-11-30
CtrlLib: WhenLeftUp event added to DocEdit/LineEdit/RichEdit

ismail-yilmaz:
Patch Introduction: New Event: WhenLeftUp

Objective:
This patch introduces the WhenLeftUp event to enhance customization options for client code.

Purpose:
The event enables developers to implement contextual actions for selected text, akin to the behavior seen in modern text editors and word processors like Microsoft Word. For instance, upon selecting text, a small menu can be triggered and displayed immediately when the left mouse button is released. Of course, it is not limited to this usage. But the real rationale behind it is this.

*Immediately available in:*
- `DocEdit`
- `LineEdit` (and its derivatives, e.g. `CodeEditor`)
- `RichEdit`

A simple example:

```
#include <CtrlLib/CtrlLib.h>

using namespace Upp;

GUI_APP_MAIN
{
	TopWindow w;
	DocEdit   e;
	e <<= "Hello world;";
	e.WhenLeftUp = [&] { if(e.IsSelection()) MenuBar::Execute([&](Bar& menu) { e.StdBar(menu); }, GetMousePos()); };
	w.Sizeable().Zoomable().SetRect(0, 0, 800, 600);
	w.Add(e.SizePos());
	w.Run();
}

```

In the above example I used `DocEdit::StdBar`, but the real purpose is to allow setting up other types of menus and "tool" bars.

This patch is proposed for the next release cycle.

Please review.


ismail-yilmaz:
Here's a visual:
output
Edit: replaced with a gif.

mirek-fidler:
I am a bit concerned about the action that LeftUp does, I mean, should we call WhenLeftUp before the action or after? Alternatively, it can replace the action somehow, but I would not go there; but I think there might be a good argument for moving it after the action. Would that work for you?

(Well, alternative view is that you can always derive a new class for this and override LeftUp, but I am quite ok with adding an Event)

ismail-yilmaz:
> I am a bit concerned about the action that LeftUp does, I mean, should we call WhenLeftUp before the action or after? Alternatively, it can replace the action somehow, but I would not go there; but I think there might be a good argument for moving it after the action. Would that work for you?
> 
> (Well, alternative view is that you can always derive a new class for this and override LeftUp, but I am quite ok with adding an Event)

Sure, if the ctrl does not lose the selection after the action (which would kinda defeat the purpose of WhenLeftUp event), it can be called after the action. Reason for the proposed event is (a) deriving the whole ctrl just to have this functionality seems to me an overkill, and (b) callbacks for such events in post-C++11 era are IMO more elegant and flexible way.

PR #222, opened 2024-12-22
Fix GetDefaultUppOut() for Flatpak

klugier:
It looks like version 2024.1 is incompatible with Flatpak. I identified that in GetDefaultUppOut() in ide/Core/Assembly.cpp we are trying to get upp out directory basing on ide executable file. In our Flatpak environment it won't work, because we are using host compiler not the sandboxed one. So, in such case we always need to base on users home directory.

### DIFF
--- a/uppsrc/ide/Core/Assembly.cpp
+++ b/uppsrc/ide/Core/Assembly.cpp
@@ -156,6 +156,8 @@ bool SaveVars(const char *name)
 String GetDefaultUppOut()
 {
 	String out;
+	
+#ifndef FLATPAK
 	String p = GetExeFolder();
 	while(p.GetCount() > 1 && DirectoryExists(p)) {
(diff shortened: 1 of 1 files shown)


mirek-fidler:
- original code places out into exe directory only if it finds .chache folder there or in any parent dirs. So the only way this is going to screw you is if there is .cache in your installation already. All of this is done to prevent "going outside of initial dir" unless user moves theide binary somewhere else... So place check that this change is needed and what you want...

- in general, I think bugfix PR should be to 'stable' branch, not master (not sure what you mean here...)?

klugier:
It looks like you are right. I had .cache directory in the place where ide executable is located inside sandbox. After proper cleaning in the Flatpak build script everything works as expected without this change. Closing and thanks for common sense approach.

PR #225, opened 2025-01-11
GridCtrl Correction and enhancement  proposition 

didierq:
- [GridCtrl::SearchImmediate(bool b = true) BUG correction](https://github.com/ultimatepp/ultimatepp/commit/41b7663eda3e602d992c5ca4f7d53c096d655530)
When calling SearchImmediate(..) in application code, it didn't work at all. Now it activates and deactivates ImmidiateSearch as expected

- [Added WhenSelectAction()](https://github.com/ultimatepp/ultimatepp/commit/87c840c36ab26e956d9fea1381cba033bae3cb3e)
This enhancement adds WhenSelectAction: WhenSelectAction is called every time row selection is modified. It allows to synchronize GridCtrl row selection with something else


didierq:
Standby on this for the moment : I just stumbled on a case: I'm not sure where it comes from but just in case.
I will do all the necessary verifications and come back to confirm or request for pullRequest delete

PR #227, opened 2025-01-25
Ide: Minmalize the possibility of scrollbar apperance in version info in about box

klugier:
This PR addressing the issue when scrollbar appears in the about box in version information due to too much content. It was achieved by decreasing font from 6 to 5. Also, small issue has been fixed. When there is no information about the branch there is no space between version and bit size of the operating system. Also, to minimize the possibility of scrollbar for version info, the new line at the end has been eliminated.

### DIFF
--- a/uppsrc/ide/About.cpp
+++ b/uppsrc/ide/About.cpp
@@ -34,13 +34,12 @@ String SplashCtrl::GenerateVersionInfo(bool qtf)
 	h << "Build: " << GenerateVersionNumber();
 #ifdef bmGIT_BRANCH
 	h << " " << bmGIT_BRANCH;
-	h << separator;
 #endif
 
 	if(sizeof(void *) == 8)
-		h << "64 bit";
+		h << " 64 bit";
 	else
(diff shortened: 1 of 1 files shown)


mirek-fidler:
I thought I fixed this one yesterday (by simply moving the text in top-right direction to make more space).

Are you sure you tested with actual master? (it is also possible I have forgot to push)

mirek-fidler:
https://github.com/ultimatepp/ultimatepp/commit/1d8ffa993e4d0234417e48bfa4edf641c77e99a2

height increased to *8 should have done the job....

klugier:
On my non high dpi display it doesn't look good. When I created this patch I was using latest master. IMO, the font should be smaler from 6 to 5 to avoid potential problems, also new line after number of classes and objects are not needed since it generates one additional line to display.

klugier:
I attach screenshot. With fix:
AboutWithFix

Without (please keep in mind that branch name in this scenario is very short - master)
AboutWithoutFixLatestMaster

PR #228, opened 2025-01-30
Ctrl geometry animation functions and docs

ismail-yilmaz:
GUI animations, whether simple or complex, are now part of the modern UI frameworks.

In this regard, this set of patches introduces:

1) A "batch" ctrl animation function:  `void Animate(Vector<Ptr<Ctrl>>& ctrls, const Vector<Rect>& targets, int duration = 100)` This function animates the transition of multiple ctrls from their current positions to the target positions specified by the targets parameter, over a given duration. 

2) A reference example `reference/AnimateCtrlGeometry`, which demonstrates the batch animation function by animating ("spreading") eight color pushers evenly in a top window.

3) Documentation of `Animate` function and its two overloads:
```
void Animate(Ctrl& c, const Rect& target, int type = -1);
void Animate(Ctrl& c, int x, int y, int cx, int cy, int type = -1);
void Animate(Vector<Ptr<Ctrl>>& ctrls, const Vector<Rect>& targets, int duration = 100);
```

Please review.


mirek-fidler:
It is good (and merged as it is).

What was wrong with Sync?

mirek-fidler:
It is good (and merged as it is).

What was wrong with Sync?

ismail-yilmaz:
> It is good (and merged as it is).
> 
> What was wrong with Sync?

On  Windows, most of the time the UI freezes till the end of animation duration or becomes very choppy..

mirek-fidler:
Tried to make it more generic (as I realised that I might need to animate something else than Ctrl as well)

ismail-yilmaz:
> Tried to make it more generic (as I realised that I might need to animate something else than Ctrl as well)

No problem.

Issue #220, opened 2024-12-17
Umk fails to download UppHub packages on Windows (git.exe not found error)


ismail-yilmaz:
Umk on windows can't download Upphub packages because umk tries to run "git" command, which may not be available on the system.

It should check and instead use (or prefer) `upp\bin\mingit\cmd\git.exe`

PR #230, opened 2025-02-01
Umk: Git exe file path fix for windows builds.

ismail-yilmaz:
This PR addresses the Git executable path issue on Windows. (See: issue #220)  

### **Problem**  
Currently, **umk** uses only `"git"` as the executable path on both POSIX systems and Windows. While this works fine on POSIX, it causes issues on Windows because the system searches for Git in system directories. If Git is not installed globally, **umk** fails to perform Git-related operations (e.g., UppHub).  

### **Solution**  
Since **mingit** is shipped by default, **umk** should first attempt to locate its executable, assuming it resides in a path relative to **umk**'s executable folder.  

This PR introduces a function: **`GetGitPath()`**, which:  
1. First checks for **mingit’s** executable.  
2. If not found, falls back to the default `"git"` executable.  

This ensures **umk** can reliably use Git on Windows.

### DIFF
--- a/uppsrc/ide/Core/Util.cpp
+++ b/uppsrc/ide/Core/Util.cpp
@@ -25,3 +25,17 @@ int GetRepoKind(const String& p)
 	String pp = p;
 	return GetRepo(pp);
 }
+
+String GetGitPath()
+{
+#ifdef PLATFORM_WIN32
+	static String path;
+	ONCELOCK {
+		path = AppendFileName(GetExeFolder(), "\\bin\\mingit\\cmd\\git.exe");
+		path = FileExists(path) ? "\"" + path + "\"" : "git";
+	}
+	return path;
+#else
+	return "git";
+#endif
+}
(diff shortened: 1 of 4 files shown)


uppsrc/ide/Core/Util.cpp
```diff
@@ -25,3 +25,17 @@ int GetRepoKind(const String& p)
...
+
+String GetGitPath()
+{
+#ifdef PLATFORM_WIN32
+	static String path;
+	ONCELOCK {
+		path = AppendFileName(GetExeFolder(), "\\bin\\mingit\\cmd\\git.exe");
+		path = FileExists(path) ? "\"" + path + "\"" : "git";
+	}
+	return path;
```
klugier, line 37: I would firstly check if git executable is available in the PATH. After that, I would look for bundled git. Alternativly, we could use diffrent order. Firstly check for mingit and then if we don't find, it we should look for os wide git executable. 

The advantage of OS wide git executable is that it is always up to date with security patches on board. We can not tell the same for git executable bundled to mingit. We do not upgrade mingit that frequent.

klugier (approved):

PR #231, opened 2025-02-07
New simd functions (GetMask variants (i8, i16, i32, f32) and equality operator (i8) for SSE2 and NEON).

ismail-yilmaz:
This PR adds some very useful and crucial code seemingly missing in U++ SIMD functions: `GetMaski8x16()`, `GetMaski16x8()`,  `GetMaski32x4()`, `GetMaskf32x4()`.

These functions allow developers to do some very useful operations easily using the SIMD instructions: Counting, accumulating and determining positions in arrays (e.g they can be used to vectorize string/byte searches).

A simple example is `reference/StreamGetSzPointer` example. The vectorized version of the example can be as follows:

```
int CountLinesOptimizedSIMD(Stream& s) {
	int n = 0;

	for (;;) {
		int sz;
		const byte* p = s.GetSzPtr(sz);

		if (sz) {
			const byte* e = p + sz;
			const byte* e16 = p + (sz & ~15);  // Process in 16-byte chunks
			i8x16 q = i8all('\n');
			while(p < e16) {
				int mask = GetMaski8x16((i8x16(p) == q));
				n += __builtin_popcount(mask); // Unfortunately, this seems to be CLANG/GCC specific.
				p += 16;
			}

			// Process remaining bytes (less than 16)
			while (p < e) {
				n += (*p++ == '\n');
			}
		}
		else {
			int c = s.Get();
			if (c < 0)
				return n;
			n += (c == '\n');
		}
	}

	return n;
}



```
The results of this operation with a ~2Gib file on a Ryzen 5600 with 16 Gib RAM (CLANG) is as follows (note that the scalar version is increased to 16 bytes chunks and 16 newline checks to give the compiler an opportunity to vectorize): 

```
CountLines(in) = 59150432
Simple 1.500799 s
CountLinesOptimized(in) = 59150432
Optimized 717.475 ms
CountLinesOptimizedSIMD(in) = 59150432
Optimized with SIMD 616.997 ms

```
Admittedly, this is not definitive, GCC can better vectorize and in fact it does. However, the aim of this example is to show how the usage pattern can be simplified using masks and equality operator for i8x16 data type.

There is also an autotest for the patch.

Please check.

### DIFF
--- a/autotest/SIMD3/Etalon.log
+++ b/autotest/SIMD3/Etalon.log
@@ -1,0 +1,11 @@
+* /home/maldoror/upp/.cache/upp.out/autotest/CLANG.Debug.Debug_Full.Shared.Sse2/SIMD3 08.02.2025 12:19:08, user: maldoror
+
+mask1 = 0xaaaa
+mask2 = 0xaa
+mask3 = 0xa
+mask4 = 0xa
+=================
+All zeros:   0x0000 (expected: 0x0000)
+All ones:    0xffff (expected: 0xffff)
+Alternating: 0xaaaa (expected: 0xaaaa)
+Single bit:  0x0020 (expected: 0x0020)
--- a/autotest/SIMD3/SIMD3.cpp
+++ b/autotest/SIMD3/SIMD3.cpp
@@ -30,7 +30,7 @@ CONSOLE_APP_MAIN
 	DUMPHEX(mask4);
 	ASSERT(mask4 == 0b1010);
 	
-	LOG("==================================");
+	LOG("=================");
 	
 	// Test 2
 	byte p[16] = {0};
@@ -57,6 +57,8 @@ CONSOLE_APP_MAIN
 	s[5] = 0x80;  // Set MSB of 6th byte
 	mask4 = GetMaski8x16(s);
 	LOG(Format("Single bit:  0x%04x (expected: 0x0020)", mask4));
+
+	CheckLogEtalon();
 	
 #else
 	Cout() << "================= NO SIMD\r\n";
--- a/autotest/SIMD3/SIMD3.cpp
+++ b/autotest/SIMD3/SIMD3.cpp
@@ -1,0 +1,64 @@
+#include <Core/Core.h>
+
+using namespace Upp;
+
+CONSOLE_APP_MAIN
+{
+	StdLogSetup(LOG_FILE|LOG_COUT);
+
+#ifdef CPU_SIMD
+
+	i8x16 a(-1, 2, -3, 4, -5, 6, -7, 8, -9, 10, -11, 12, -13, 14, -15, 16);
+	i16x8 b(-1, 2, -3, 4, -5, 6, -7, 8 );
+	i32x4 c(-1, 2, -3, 4);
+	f32x4 d(-1.0f, 2.0f, -3.0f, 4.0f);
+
+	// Test 1
+	auto mask1 = GetMaski8x16(a);
+	DUMPHEX(mask1);
+	ASSERT(mask1 == 0b1010101010101010);
+
+	auto mask2 = GetMaski16x8(b);
+	DUMPHEX(mask2);
+	ASSERT(mask2 == 0b10101010);
+
+	auto mask3 = GetMaski32x4(c);
+	DUMPHEX(mask3);
+	ASSERT(mask3 == 0b1010);
+	
+	auto mask4 = GetMaskf32x4(d);
+	DUMPHEX(mask4);
+	ASSERT(mask4 == 0b1010);
+	
+	LOG("==================================");
+	
+	// Test 2
+	byte p[16] = {0};
+	mask1 = GetMaski8x16(p);
+	// Case 1: All zeros
+	LOG(Format("All zeros:   0x%04x (expected: 0x0000)", mask1));
+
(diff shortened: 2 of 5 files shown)


mirek-fidler:
https://learn.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=msvc-170

I think we need to have that in Ops.h (CountBits / CountBits64 ?)

It seems to me like having GetMaski8x16 is enough - you can always mask out uneeded bits.

Let us try to find better NEON...

mirek-fidler:
https://community.arm.com/arm-community-blogs/b/servers-and-cloud-computing-blog/posts/porting-x86-vector-bitmask-optimizations-to-arm-neon

The "bad" variant still seems much better that what is in the patch... :)

ismail-yilmaz:
@mirek-fidler 

> I think we need to have that in Ops.h (CountBits / CountBits64 ?)

OK, I'll open a separate pull request for that. But here's a [preview](https://godbolt.org/z/jYqKj14x3) on godbolt.
`std::bitset` version is more portable and on GCC/CLANG it produces virtually the same code as the builtin versions. 
Difference is also not great on MSVC but the `CountBitsBuiltin` and  `CountBitsBuiltin64` is a bit faster in some runs. But I guess it is negligable. The "fallback" code is essentially the same as builtin ones, but  with no call.

Google benchmark from my local machine (Linux):

```
Run on (12 X 4464 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 512 KiB (x6)
  L3 Unified 16384 KiB (x1)
Load Average: 0.21, 0.23, 0.33
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
CountBits               0.228 ns        0.228 ns   2963148792
CountBits64             0.226 ns        0.226 ns   3095705231
StdBitsCountBits        0.226 ns        0.226 ns   3081036403
StdBitsCountBits64      0.226 ns        0.226 ns   3083270739
```



> It seems to me like having GetMaski8x16 is enough - you can always mask out uneeded bits.

Yes, but I was thinking more about code clarity there. We can manually manipulate bits, but it will make the code less readable and more error prone in some situations. 

> Let us try to find better NEON...

Yeah, finding a counterpart for movemask on NEON is "hard". Will try harder. :)

mirek-fidler:
(one last comment wrt upptst I forgot yesterday: you might want to check out the etalon testing... e.g. upptst/MakeString)

mirek-fidler:
I think we need to make result of "GetMask" operation implementation defined and add a bunch of followup functions

auto x = GetMask(xxx);
IsMaskAny16x8(x)
IsMaskAll16x8(x)
GetFirst16x8(x)
GetLast16x8(x)
GetCount16x8(x)
Get16x8(x, i)

to solve NEON issue. (With current implementation it does not really make sense to add this at all - NEON would be slower than non-SIMD code so that defeats the whole purpose of making SIMD code dual-platform)

(I do not really like proposed names above, work in progress)

mirek-fidler:
Or actually skip GetMask completely (at interface level) and add

IsTrueAny16x8(x)
IsTrueAll16x8(x)
GetFirstTrue16x8(x)
GetLastTrue16x8(x)
GetTrueCount16x8(x)
GetTrue16x8(x, i)

(and then 8x16 and 4x32 variants)

mirek-fidler:
Little test:

#include <Core/Core.h>

using namespace Upp;

force_inline bool   IsTrue(f32x4 a, int i)        { return _mm_movemask_ps(a.data) & (1 << i); }

f32x4 a(1, 2, 3, 4);
f32x4 b(0, 2, 3, 0);

int n = 0;

CONSOLE_APP_MAIN
{
	auto x = a == b;
	for(int i = 0; i < 4; i++)
		if(IsTrue(x, i))
			n++;
	RDUMP(n);
}

I wanted to test that compiler will not do _mm_movemask_ps in every iteration. Confirmed....

mirek-fidler:
All done. Looking at disassembly, 32bit arm works just fine with 64bit mask, so I have left the NEON code the same.

https://github.com/ultimatepp/ultimatepp/commit/2ae4a30077785423a3a40fde82e905946a19e943

Feel free to comment and/or close the request...

mirek-fidler:
Also added reference example. Interestingly, CLANG happily autovectorizes non SSE variant, albeit with little bit clunky code. Anyway:

CountLines(in) = 4048112
Simple 802.187 ms
CountLinesOptimized(in) = 4048112
Optimized 578.223 ms
CountLinesOptimizedSIMD(in) = 4048112
Optimized SIMD 515.738 ms

ismail-yilmaz:
> All done. Looking at disassembly, 32bit arm works just fine with 64bit mask, so I have left the NEON code the same.
> 
> [2ae4a30](https://github.com/ultimatepp/ultimatepp/commit/2ae4a30077785423a3a40fde82e905946a19e943)
> 
> Feel free to comment and/or close the request...

Hello Mirek,

Thank you very much for this work!  I'll happily close the request. Also I'll check the code this weekend and -hopefully- confirm that it's all good. 

Best regards,

PR #232, opened 2025-02-08
Bit counting functions (CountBits, CountBits64)

ismail-yilmaz:
This PR adds `CountBits()` and `CountBits64()` functions to `Upp/Core` and provides API docs & a simple autotest.

These functions are useful for counting set bits in a bitmask.

Default implementation uses platform specific popcount variants for popular compilers (GCC/CLANG/MSVC). Otherwise it uses a well-known custom implementation for fallback, which can be easily vectorized by the compiler as well.

 Please check.

### DIFF
--- a/autotest/CountBits/CountBits.cpp
+++ b/autotest/CountBits/CountBits.cpp
@@ -1,0 +1,33 @@
+#include <Core/Core.h>
+
+using namespace Upp;
+
+CONSOLE_APP_MAIN
+{
+	StdLogSetup(LOG_FILE|LOG_COUT);
+	
+	DUMP(CountBits(0x0001));
+	DUMP(CountBits(0x000f));
+	DUMP(CountBits(0x00ff));
+	DUMP(CountBits(0x0fff));
+	DUMP(CountBits(0xffff));
+	DUMP(CountBits(0x000fffff));
+	DUMP(CountBits(0x00ffffff));
+	DUMP(CountBits(0x0fffffff));
+	DUMP(CountBits(0xffffffff));
+	
+	
+	LOG("====================");
+	
+	DUMP(CountBits64(0x00000001ffffffff));
+	DUMP(CountBits64(0x0000000fffffffff));
+	DUMP(CountBits64(0x000000ffffffffff));
+	DUMP(CountBits64(0x00000fffffffffff));
+	DUMP(CountBits64(0x0000ffffffffffff));
+	DUMP(CountBits64(0x000fffffffffffff));
+	DUMP(CountBits64(0x00ffffffffffffff));
+	DUMP(CountBits64(0x0fffffffffffffff));
+	DUMP(CountBits64(0xffffffffffffffff));
+	
+	CheckLogEtalon();
+}
--- a/autotest/CountBits/CountBits.upp
+++ b/autotest/CountBits/CountBits.upp
@@ -1,0 +1,10 @@
+uses
+	Core;
+
+file
+	Etalon.log,
+	CountBits.cpp;
+
+mainconfig
+	"" = "";
+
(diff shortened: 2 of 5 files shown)


mirek-fidler:
It is perfect, thanks

ismail-yilmaz:
@mirek-fidler 

Out of curiosity: Is this going to get merged after 2025.1 release or since you're very busy, did you just forget to merge it? :)

mirek-fidler:
Ha, thanks. I thought it is merged..

kov-serg:
Why this function is not covered with option to disable this feture and use fallback.
Something like:
```
#if ( defined(__GNUC__) || defined(__clang__) ) && defined(CPU_SSE42)
    return __builtin_popcount(mask);
#elif defined(_MSC_VER) && defined(CPU_SSE42)
    return __popcnt(mask);
...
```
Is it hard to do it propper way? Compiler generate popcnt instruction but target may not support it. Especially if target is 32bit.

https://stackoverflow.com/questions/39670781/how-do-i-detect-whether-the-cpu-has-good-popcnt-support

mirek-fidler:
OK, so this is a good point to consider.

It looks like __builtin_popcount should work even if not supported by ABM (I do not have any CPU to check though...)

Which leaves us with MSC version. POPCNT should be avialable since Haswell, which accidentally is also minimal CPU to run Win11. So we are solving here situation where you want to build for unsupported windows version with MSC compiler, am I correct?

But I guess we CAN add something like flagNOSSE3 or something...

mirek-fidler:
Hm, but this seems quite useful discussion, testing with clang in windows, this is what is produced

716B13E0  mov eax,[rel 0x71c6f000] 
716B13E6  mov ecx,eax 
716B13E8  shr ecx,1 
716B13EA  and ecx,0x55555555 
716B13F0  sub eax,ecx 
716B13F2  mov ecx,eax 
716B13F4  and ecx,0x33333333 
716B13FA  shr eax,0x2 
716B13FD  and eax,0x33333333 
716B1402  add eax,ecx 
716B1404  mov ecx,eax 
716B1406  shr ecx,0x4 
716B1409  add ecx,eax 
716B140B  and ecx,0xf0f0f0f 
716B1411  imul eax,ecx,dword 0x1010101 
716B1417  shr eax,0x18 
716B141A  mov [rel 0x71c6f000],eax 
716B1420  ret 

so it means you need to activate SSE4 or something in the compiler anyway...

mirek-fidler:
clang needs -mpopcnt ...

016213E0  popcnt eax,dword [rel 0x1bdf000] 
016213E8  mov [rel 0x1bdf000],eax 
016213EE  ret 

And the lowest CPU is in fact 2008 Nehalem....

mirek-fidler:
For now my conclusion is

- we need to add -mpopcnt to default build methods (maybe just windows?)
- we need flagLEGACY_CPU that would in general activate support for "CPUs older than 15 years". Which OTOH is a good chance to start using avx2...

kov-serg:
You can add it by default. But there must be way to change it from one place.
There are a lot of unique retro computers hardware.
I use gcc for winxp target :) https://github.com/kov-serg/ultimatepp/releases/tag/2022.1_xp

ps: you are older than "some CPUs", is it a reason to give up. Something must works longer than humans lifetime. Now there are a lot of instant disposable software that can work only here and now, for a short period of time.

mirek-fidler:
OK, so -mpopcnt should now be default with x86 and we have a new flag "LEGACY_CPU"... (I am still undecided whether to support CPUs without SSE2, to stay on the safe side I am disabling SSE2 if LEGACY_CPU is in main config).

Please test.

P.S.: Unfortunately, I am older than 4004 :)

kov-serg:
Very good.
First thing is to fix Debug.cpp
https://github.com/ultimatepp/ultimatepp/blob/master/uppsrc/Core/Debug.cpp#L202
```
#ifndef flagNOSIMD
String AsString(__m128i x)
...
#endif
```

ps: I think this pattern better than simple hardcoded 256
```
enum { buf_size=256 }; char buf[buf_size];
...
snprintf(buf,buf_size, ...
```

pps: "Please test" -- unable to test it today. It wants ide/clang but gcc . So I have to figure out how to turn it off.  I'll test it next week.

PR #233, opened 2025-02-09
Ask user to manually install Homebrew

9t8:
Homebrew now needs the user to add it to the path manually. As such, the script no longer works as-is. This change makes the script more robust.

### DIFF
--- a/uppbox/MakePosixInstall/scripts.cpp
+++ b/uppbox/MakePosixInstall/scripts.cpp
@@ -5,7 +5,7 @@ AskContinue()
 {
   read -p "Continue (Y/n)?" answer
   if [ "$answer" != "${answer//[nN]/x}" ]; then
-	 exit;
+    exit;
   fi
 }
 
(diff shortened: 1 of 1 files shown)


klugier:
Can you provide a link to the documentation page where the authors of Homebrew stands that the default behavior has been changed and there is a need to manually modify PATH? Or alternatively, commit with the change. Also, I didn't see any instructions on https://brew.sh/ that PATH must be configure manually. Maybe you hit a bug and wants to change behavior for everyone?

klugier (changes requested):
Please provide profs that the previous default behavior has been changed. I look at brew official page and it doesn't say nothing about the need for adding homebrew to the path after installation.

9t8:
The code asking the user to change dotfiles manually is [here](https://github.com/Homebrew/install/blob/master/install.sh#L1047-L1088). You can read the script to find it never modifies PATH.

klugier:
OK, I found this commit https://github.com/Homebrew/install/commit/32155c0ea474a43dc6af31d940ec0f818fdd7827. It looks like on ARM64 (M CPU families) you need to add brew to PATH by your own. OK, I am fine with these changes.

klugier (approved):

uppbox/MakePosixInstall/scripts.cpp
```diff
@@ -14,11 +14,12 @@ uname=`uname`
...
   DEP=""
   if which brew; then
-    echo "Homebrew is already installed"
+    echo "Homebrew is installed"
   else
-    echo "Installing Homebrew package management system (to obtain openssl libraries required for U++)"
+    echo "Homebrew is not installed"
+    echo "We use Homebrew to obtain OpenSSL libraries and clang-format required for U++"
+    echo "Follow the instructions at https://brew.sh/ to install Homebrew"
     AskContinue
```
klugier: We should exit instead of AskContinue. You had to install brew prior to launching this script, so I would just replace this line with exit 1 or something like that.

Theoretically, you can install brew in the second tab of your terminal, but it is difficult to predict whenever in this place it will be update. So, it is much more safer to end script execution in this place.

PR #235, opened 2025-02-18
Core: Lerp overload with custom easing function.

ismail-yilmaz:
This PR does two things:

1) Makes the existing Lerp function `constexpr` and marks it with `noexcept`
2) Proposes a Lerp overload that can take a custom easing function. This  easing function can easily be used to achieve different effects, such as bounce-in/bounce-out, ease-in/ease-out, etc.

An example easing function (bounce out) can look like this:

```
constexpr double BounceOut(double t) noexcept // Piecewise quadratic function to create a bouncing effect.
{
	if(t < 4.0 / 11.0)
		return (121.0 * t * t) / 16.0;
	if(t < 8.0 / 11.0)
		return (363.0 / 40.0 * t * t) - (99.0 / 10.0 * t) + 17.0 / 5.0;
	if(t < 9.0 / 10.0)
		return (4356.0 / 361.0 * t * t) - (35442.0 / 1805.0 * t) + 16061.0 / 1805.0;
	return (54.0 / 5.0 * t * t) - (513.0 / 25.0 * t) + 268.0 / 25.0;
}

```

This patch does and should not break anything.

P.s.: While the C++17 standard implicity assumes that a constexpr function is hinting inline, I explicitly added them for clarity.

Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -428,6 +428,7 @@ $(OutDir_ide)BaseDlg.o: $(UPPDIR1)ide/BaseDlg.cpp \
 	$(UPPDIR1)Core/CallbackN.i \
 	$(UPPDIR1)Core/CallbackNP.i \
 	$(UPPDIR1)Core/CallbackR.i \
+	$(UPPDIR1)Core/CharFilter.h \
 	$(UPPDIR1)Core/CharSet.h \
 	$(UPPDIR1)Core/CharSet.i \
 	$(UPPDIR1)Core/CoAlgo.h \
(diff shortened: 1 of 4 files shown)


mirek-fidler:
Are you sure you have commited the correct branch here? Files IMO do not correspond with description....

ismail-yilmaz:
Oo

> Are you sure you have commited the correct branch here? Files IMO do not correspond with description....

Oh, there must be a glitch in the matrix. I'm closing this and will re-open a new version.

PR #241, opened 2025-03-17
Tutorial(Gui24): make text readable under dark theme

BLumia:
Make text readable under dark theme by changing the text color to `SBlue`.

Before the patch:

image

After the patch:

image

Also, `zh-cn` is Simplified Chinese instead of Traditional Chinese (which is `zh-tw` or `zh-hk`). I can submit another PR to correct the language issue.

By the way: Does TheIDE itself support localization? How can I change the UI language of TheIDE to, for example, Chinese? I can't find related option across the IDE settings dialog, did I miss something?

### DIFF
--- a/tutorial/Gui24/main.cpp
+++ b/tutorial/Gui24/main.cpp
@@ -13,7 +13,7 @@ struct MyAppWindow : TopWindow {
     
     void Paint(Draw& w) override {
         w.DrawRect(GetSize(), SLtYellow);
-        w.DrawText(20, 20, t_("Hello translation engine!"), Arial(30), Blue);
+        w.DrawText(20, 20, t_("Hello translation engine!"), Arial(30), SBlue);
     }
 };
 


mirek-fidler:
Thank you. Correcting chinesse would be great, we lack knowledge there. ("tw" and "hk" seem like "taiwan" and "honkkong". Is it the correct flavor of chinesse?).

TheIDE does not support localisation and we do not plan to do so: Even if some heroic effort we would be able to localise it now, it would mean unsustainable amount of work in future. Meanwhile, I seriously doubt that you can use C++ without basic knowledge of English...

BLumia:
> ("tw" and "hk" seem like "taiwan" and "honkkong". Is it the correct flavor of chinesse?)

Correct. TW and HK both use Traditional Chinese, while sometimes there are subtle differences when using certain words.

> TheIDE does not support localisation and we do not plan to do so: Even if some heroic effort we would be able to localise it now, it would mean unsustainable amount of work in future.

Understandable, while I believe the community might be able to help, I'm personally interested in helping if it's desired, but feel free to keep it as English-only if preferred.

> Meanwhile, I seriously doubt that you can use C++ without basic knowledge of English...

I personally can use TheIDE without issue but there are people who not actually good at English but can code. For example, for education purpose, students might only need to know how the C/C++ grammar works so they can code. Their teaching material will likely be the same language as their mothertongue. As a GUI application, if users are familiar with a concept, they can try to find the matching word on the UI. If they are only familiar with the names that was taught in their mothertongue instead of the name in English, they might need to do the related resource before actually getting their job done.

I actually agree actual C++ developers need to have basic knowledge of English btw :)

---

I'll submit the PR that correct the language in Gui24 example next day :)

PR #242, opened 2025-03-17
Core: ProcessHasAdminRights() function is added to utility functions.

ismail-yilmaz:
Sometimes it is very useful to know if the current process/app is running in administrator/root mode. In this way, developers can enable or disable certain set of features or warn the user about the security risk posed by escalated privileges.

This PR,

- Adds `ProcessHasAdminRights()` function to U++, which can check if the current process is running in admin (Win) or root (POSIX) mode.  
- Adds the relevant example.
- Adds the relevant API doc entry.

Tested on Linux, Win 10/11.

Please review.

### DIFF
--- a/reference/DetectPrivilegedProcess/main.cpp
+++ b/reference/DetectPrivilegedProcess/main.cpp
@@ -5,7 +5,7 @@ using namespace Upp;
 GUI_APP_MAIN
 {
 	// Run the example as administrator (Windows) or as root (POSIX) to get the warning.
-	if(IsAdmin())
+	if(IsUserAdmin())
 		Exclamation(t_("Warning: Application has administrator/root rights.&This might pose a security risk!"));
 	else
 		PromptOK(t_("Application has user rights."));
--- a/reference/DetectPrivilegedProcess/main.cpp
+++ b/reference/DetectPrivilegedProcess/main.cpp
@@ -5,7 +5,7 @@ using namespace Upp;
 GUI_APP_MAIN
 {
 	// Run the example as administrator (Windows) or as root (POSIX) to get the warning.
-	if(ProcessHasAdminRights())
+	if(IsAdmin())
 		Exclamation(t_("Warning: Application has administrator/root rights.&This might pose a security risk!"));
 	else
 		PromptOK(t_("Application has user rights."));
--- a/reference/DetectPrivilegedProcess/main.cpp
+++ b/reference/DetectPrivilegedProcess/main.cpp
@@ -1,0 +1,12 @@
+#include <CtrlLib/CtrlLib.h>
+
+using namespace Upp;
+
+GUI_APP_MAIN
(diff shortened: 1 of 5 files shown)


reference/DetectPrivilegedProcess/main.cpp
```diff
@@ -0,0 +1,12 @@
+#include <CtrlLib/CtrlLib.h>
+
+using namespace Upp;
+
+GUI_APP_MAIN
+{
+	// Run the example as administrator (Windows) or as root (POSIX) to get the warning.
+	if(ProcessHasAdminRights())
```
klugier: IMO, More convient is HasProcessAdminRights(). However, not sure if adding a function is the options here. I saw that Apple for example has ProcessInfo class, maybe we should have the same with more info available to the library consumer?

IMO, It is easy to add new function, but making it consistent with the rest of the framework is not trivial.

ismail-yilmaz: Thanks for the review.

> I saw that Apple for example has ProcessInfo class, maybe we should have the same with more info available to the library consumer?

This is an option. But I think it might be hard to maintain cross-platform consistency for such a class/struct. Each platform has it own set of attributes, etc. So it might be the subject of another PR. 

OTOH, the proposed function here is simple, does what it describes and consistently works on all major platforms (Linux,BSD, MacOS, Windows), as it uses the POSIX-compliant `geteuid()` sys call on the first three, and on Windows it actually supports back to WinXP.

Function name can be changed of course (e.g. `IsRoot()`, or `IsAdmin()` or something else), and if we implement some ProcessInfo class later the implementation can still be changed to reflect it (but it will be an implementation detail, the result should be effectively the same).

User0755: Hi,

I'm personally all for this IsRoot() / IsAdmin() approach. The shorter the better: It is also easier to remember or find when needed by just typing "Is", activate Assist and then pick from the list. (If some more complex process info class is desired by someone, by all means, go for it... But please don't drop this idea.)

Best regards,

Tom

ismail-yilmaz: OK, I've renamed the function as `IsAdmin()`.  This rings better.

ismail-yilmaz: @mirek-fidler  

Any thoughts about this one?

mirek-fidler: I like the concept. IsAdmin feels touch too short for global function.

IsUserAdmin ? :)

ismail-yilmaz: > I like the concept. IsAdmin feels touch too short for global function.
> 
> IsUserAdmin ? :)

OK, I renamed it as `IsUserAdmin()`.

PR #244, opened 2025-03-18
RichEdit: Fix issue with unresponsive menu bar after right click on Wayland

klugier:
It seems that executing LeftDown when text is not slected in RightDawn in RichEdit doesn't make much sense. After removing this line the Wayland version of RichEdit works as expected and right click menu bar when there is no selection behaves as expected.

Not sure if there is a drawback of removing of LeftDawn, hoever I didn't notice anything bad.


mirek-fidler:
LeftDown is there to move cursor to right-click location.

It is common practice used over many widgets (and user code). If that hangs for some reason, it has to be fixed in CtrlCore.

klugier:
Thanks for the tip. I will try to find a root cause then.

PR #251, opened 2025-04-07
Color Added

XKFS:
More colors added and also added to the tpp.


ismail-yilmaz:
I have a suggestion: AFAIK the original 16 (8/8) color palette has over 60 years of history, so I think it might be better to put the new ones after the original 16 [8/8] colors, in both code and docs, as a separate group (as "additional" or "extra" colors maybe?), for the sake of clarity.

mirek-fidler:
I am not sure either way, but how do we feel about SOrange() ? Or SDkRed()? (later is probably basically pink in dark mode...)

mirek-fidler:
I dislike the change to LtYellow. Now your definition might even be better, but this would break existining code.

PR #262, opened 2025-04-23
FileSelNative: Fixed gtk variant to allow for multiple file types

RealCaptainBerty:
Fixed the GTK variant of the `FileSelNative` class to allow for multiple file types in a single filter (as in Win32).
For example:
```
    FileSelNative sel;
    sel.Type("File type 1 or type 2", "*.typ1;*.typ2");
```


mirek-fidler:
As FileSel is using space for separation of masks, can you try that this works for you

Vector<String> patterns = Split(type[i].b, [](int c) { return c == ';' || c == ' '; });

?

RealCaptainBerty:
Yes, works fine!

mirek-fidler:
Thank you!

PR #270, opened 2025-05-18
Core/SSL: Secure Functions for Cryptographic Random/Nonce Generation

ismail-yilmaz:
## Rationale
U++ currently lacks a dedicated cryptographically secure nonce generator suitable for modern cryptographic protocols. This pull request introduces thread-safe nonce generation functions to the `Core/SSL` package, providing guaranteed uniqueness and fork-safety for cryptographic operations.

## Features
* **Guaranteed uniqueness** through process-unique identifiers and atomic counters
* **Cryptographic randomness** via OpenSSL's CSPRNG
* **Fork-safe** on POSIX systems with automatic state regeneration
* **Two generation modes**:
  - Structured nonces (process UID + counter + optional random tail)
  - Pure cryptographic randomness
* **Standard-compliant** sizes for common protocols (AES-GCM, ChaCha20-Poly1305, JWT, OCSP)
* **Thread-safe** through lock-free atomic operations and lightweight SpinLock
* **Type-safe** return values using `SecureBuffer<byte>` with automatic memory zeroing

## Implementation Details
The generator combines three entropy sources:
1. **Process-unique identifier** (64-bit CSPRNG-generated UID)
2. **Atomic counter** with overflow protection
3. **OpenSSL CSPRNG** for random tail padding

Nonce structure adapts based on size:
- **12-15 bytes**: `[4B UID | 8B counter | random tail]`
- **16+ bytes**: `[8B UID | 8B counter | random tail]`

## Public API
```cpp
// Core functions
SecureBuffer<byte> SecureRandom(int n);  // Pure CSPRNG output
SecureBuffer<byte> SecureNonce(int n);   // Structured nonce (min 12 bytes)

// Protocol-specific helpers
SecureBuffer<byte> GetAESGCMNonce();           // 12 bytes, AES-GCM/ChaCha20-Poly1305
SecureBuffer<byte> GetChaChaPoly1305Nonce();   // 12 bytes, alias for AES-GCM
SecureBuffer<byte> GetTLSNonce();              // 12 bytes, TLS 1.2/1.3
SecureBuffer<byte> GetAESCCMNonce();           // 13 bytes, AES-CCM
SecureBuffer<byte> GetJWTNonce();              // 16 bytes, JWT tokens
SecureBuffer<byte> GetOAuthNonce();            // 16 bytes, OAuth flows
SecureBuffer<byte> GetOCSPNonce();             // 20 bytes, OCSP requests
[...]

PR #272, opened 2025-05-31
Core/SSL: Add `SecureBuffer` and `SecureZero` for Secure Cryptographic Data Handling

ismail-yilmaz:
## Rationale

This PR introduces `SecureBuffer<T>` class and `SecureZero()` function(s) for handling sensitive cryptographic data in memory, which U++ currently lacks. These utilities provide memory locking and cryptographic erasure to reduce the risk of sensitive data being swapped to disk or left in memory after use.
 
## Features

-   **`SecureZero<T>()`**:
    
    -   Safely and securely zeroes memory, using `OPENSSL_cleanse()`.
        
    -   Safely zeroes static arrays.
        
    -   Enforces that `T` is trivially copyable and destructible via `static_assert`.
        
-   **`SecureBuffer<T>`**:
    
    -   RAII-managed dynamic buffer for sensitive data.
        
    -   Automatically zeroes memory on destruction using `SecureZero` on destruction.
    -   Prevents paging sensitive memory using `mlock` or `VirtualLock` when possible ("best-effort" locking).
    -  Pick-only semantics to prevent accidental copies.
    -  Enforces that `T` is trivially copyable and destructible via `static_assert`.     

## Format

`SecureBuffer<T>` behaves like a dynamic container (`T*`, `begin()`, `end()`, `operator[]`), very similar to `Upp::Buffer<T>`, but is not intended for general-purpose use. It is specialized for securely storing sensitive data.

## Public API

```cpp
// SecureZero
template <class T>
void SecureZero(T* ptr, size_t count);

template<class T, size_t N>
void SecureZero(T (&obj)[N]); // static array overload

// SecureBuffer
template <class T>
class SecureBuffer : Moveable<SecureBuffer<T>>, NoCopy {
public:
    SecureBuffer()
    SecureBuffer(size_t size_)
    SecureBuffer(SecureBuffer&& src)
    ~SecureBuffer()

    SecureBuffer& operator=(SecureBuffer&& src)

    void        Alloc(size_t size)
    void        Clear()
    void        Zero()

    size_t      GetSize() const

    bool        IsEmpty() const

    operator T*()
    operator const T*() const
    T *operator~()
    const T *operator~() const
    T          *Get()
    const T    *Get() const
    T          *begin()
    const T    *begin() const

    T*          Begin()
    const T*    Begin() const


    T*          end()
    const T*    end() const
    T*          End()
    const T*    End() const


    T& operator[](size_t i)
    const T& operator[](size_t i) const
};
 
```

### Notes:
 - A constructor with `std::initializer_list` is intentionally left out of the design, as it can be dangerous in cryptographic contexts.

## Unit Tests

-   SecureZero:
    
    - Basic integer array
    - Empty array (edge case)
    - NULL pointer (edge case)
    - Byte-by-byte zeroing
    - Complex types
    - Partial array zeroing
    - Resistance to compiler optimization
    - Overlapping memory regions
        
        
-   SecureBuffer:
    
     - Basic functionality
     - Pick semantics
     - Zeroing verification (security critical)
     - Multiple clear calls
     - Edge case

## API Docs

-   Included.
   
## Changes
- None. This is a new addition to `Core/SSL` package.

## Related Discussions

#269 
#270

### DIFF
--- a/autotest/SecureBuffer/SecureBuffer.cpp
+++ b/autotest/SecureBuffer/SecureBuffer.cpp
@@ -1,0 +1,148 @@
+#include <Core/Core.h>
+#include <Core/SSL/SSL.h>
+
+using namespace Upp;
+
+template<typename T>
+bool IsZeroed(const T* ptr, size_t count)
+{
+	auto bytes = reinterpret_cast<const byte*>(ptr);
+	for(size_t i = 0; i < count * sizeof(T); ++i) {
+		if(bytes[i] != 0) {
+			LOG(Format("Memory not zeroed at offset %d: found value 0x%02X", (int) i, (int) bytes[i]));
+			return false;
+		}
+	}
+	return true;
+}
+
+CONSOLE_APP_MAIN
+{
+	StdLogSetup(LOG_COUT | LOG_FILE);
+	
+	auto Test = [](const String& name, const Function<void()>& fn) {
+		String txt = "---" + name + ": ";
+		fn();
+		LOG(txt << "PASSED");
+	};
+
+	Test("SecureZero: Basic integer array", [&]{
+		int buffer[10] = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
+		SecureZero(buffer);
+		ASSERT(IsZeroed(buffer, 10));
+	});
+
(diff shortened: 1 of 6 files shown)


mirek-fidler:
Excellent work, thank you!

PR #283, opened 2025-06-15
Added reset button to reference/AnimateCtrlGeometry plus changes in parameter name (duration -> duration_ms)

klugier:
The changes introduced in this PR improves CtrlLib animations by:
- Adding reset button to reference/AnimateCtrlGeometry, it was annoying to relaunch the reference app to see the animation again
- Change the animate function parameter name duration to duration_ms. Thanks to that the signature should be easier to understand, also it is aligned with SetTimeCallback declaration.
- Docs has been updated accordingly to changes in Animate function parameter names

### DIFF
--- a/reference/AnimateCtrlGeometry/main.cpp
+++ b/reference/AnimateCtrlGeometry/main.cpp
@@ -5,26 +5,50 @@ using namespace Upp;
 struct AnimateCtrlGeometry : TopWindow {
 	Array<ColorPusher> pushers;
 	Vector<Ptr<Ctrl>>  ctrls;
+	Rect               ctrls_start_rect;
 	Vector<Rect>       targets;
-
+	Button             reset;
+	
 	AnimateCtrlGeometry()
+		: ctrls_start_rect(Rect(0, 0, 100, 100))
 	{
 		Title("Ctrl geometry animation");
 		
-		CenterScreen().SetRect(0,0, 400, 200);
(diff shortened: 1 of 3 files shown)


ismail-yilmaz:
Thanks!

I see no problem, aside from the `AddResetButton()` function, which I find a bit unnecessary for adding a single button. :)

reference/AnimateCtrlGeometry/main.cpp
```diff
@@ -24,7 +26,33 @@ struct AnimateCtrlGeometry : TopWindow {
...
+		constexpr int launch_delay_ms = 1000;
+		constexpr int duration_ms = 300;
+		
+		reset.Disable();
+		SetTimeCallback(launch_delay_ms, [=]{
+			Animate(ctrls, targets, duration_ms); reset.Enable();
+		});
+	}
+	
+	void AddResetButton()
```
ismail-yilmaz: Isn't this a bit overkill for adding a single button? Nothing deal breaking, of course.

klugier:
OK, I removed `AddResetBetton()` and move the logic to constructor. Thanks!

ismail-yilmaz (approved):

reference/AnimateCtrlGeometry/main.cpp
klugier: Fixed! Thanks!

PR #286, opened 2025-06-29
CodeEditor: Remove "thread" as C++ keyword

klugier:
It looks like thread is not c++ keyword. Leaving it as it is might be confusing especially when you are using std::thread.

std::thread shouldn't be higlighted, in current implementation it is.

Source: https://en.cppreference.com/w/cpp/keywords.html


mirek-fidler:
Thanks.

PR #291, opened 2025-07-05
Fix warning with lambda captuer mplicit capture of 'this' with a capture default of '=' for Gui01 (c++20 and above)

klugier:
It looks like fixing the issue with deprecation of capturing this with [=] won't be that difficult. I just wanted to start the whole process. I fixed this warning for Gui01 tutorail and associated pacakges, so right now while compiling Gui01 example with std=c++20 and above there shouldn't be following warning:
`C:\Prototable\upp\git\uppsrc\CtrlLib\CtrlUtil.cpp (410): warning: implicit capture of 'this' with a capture default of '=' is deprecated [-Wdeprecated-this-capture]`

Also, I think the code is batter, because we are capturing variables we don't want to. I suggest making this kind of changes in chunks. At some point of time, we will eliminate all warnings of that kind.

I would like to add that maybe we might do not like the new syntax and it is longer to what we had in the past. But, the overall solution is batter, especially when you capture [this] instead of [=]. In this case the local variables are not copied. In the end we shouldn't consider silencing this warning. Also, at some point it might be a compilation error, so we will be forced to fix it.

--------------------------------
The next step would be to fix all Gui* tutorials. After that we can jump to other places like ide.

### DIFF
--- a/uppsrc/Core/CoWork.h
+++ b/uppsrc/Core/CoWork.h
@@ -183,7 +183,7 @@ class AsyncWork {
 		Ret2   ret;
 	
 		template<class Function, class... Args>
-		void        Do(Function&& f, Args&&... args) { co.Do([=]() { ret = f(args...); }); }
+		void        Do(Function&& f, Args&&... args) { co.Do([=, this]() { ret = f(args...); }); }
 		const Ret2& Get()                            { return ret; }
 		Ret2        Pick()                           { return pick(ret); }
 	};
--- a/uppsrc/Core/CoWork.h
+++ b/uppsrc/Core/CoWork.h
@@ -183,7 +183,7 @@ class AsyncWork {
 		Ret2   ret;
 	
 		template<class Function, class... Args>
-		void        Do(Function&& f, Args&&... args) { co.Do([=, this]() { ret = f(args...); }); }
+		void        Do(Function&& f, Args&&... args) { co.Do([=, *this]() { ret = f(args...); }); }
 		const Ret2& Get()                            { return ret; }
 		Ret2        Pick()                           { return pick(ret); }
 	};
--- a/uppsrc/Core/FilterStream.h
+++ b/uppsrc/Core/FilterStream.h
@@ -36,7 +36,7 @@ public:
 	void Set(Stream& in_, F& filter) {
 		Init();
 		in = &in_;
-		filter.WhenOut = [=](const void *ptr, int size) { Out(ptr, size); };
+		filter.WhenOut = [this](const void *ptr, int size) { Out(ptr, size); };
 		Filter = [&filter](const void *ptr, int size) { filter.Put(ptr, size); };
 		End = [&filter] { filter.End(); };
 	}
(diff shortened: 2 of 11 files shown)


klugier:
OK, I see where is the problem. When switch backed to c++17:
`uppsrc/Core/CoWork.h (186): warning: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Wc++20-extensions]`

It looks like it was in only one place. For the rest I don't use [=, this], instead I use more specific targeting. The only places that needed to be changed is AsyncWork::Impl. Now it is:
`		void        Do(Function&& f, Args&&... args) { co.Do([=, *this]() { ret = f(args...); }); }`

ismail-yilmaz:
> OK, I see where is the problem. When switch backed to c++17: `uppsrc/Core/CoWork.h (186): warning: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Wc++20-extensions]`
> 
> It looks like it was in only one place. For the rest I don't use [=, this], instead I use more specific targeting. The only places that needed to be changed is AsyncWork::Impl. Now it is: ` void Do(Function&& f, Args&&... args) { co.Do([=, *this]() { ret = f(args...); }); }`

@klugier 
Well, this should work if we don't want implicit copy, but it is too verbose and IMO, very ugly. :)

```
template<class Function, class... Args> 
void Do(Function&& f, Args&&... args) {  
        co.Do([this, f = std::forward<Function>(f), args = std::forward_as_tuple(args...)]() { 
        ret = std::apply(f, args); 
    }); 
}
```

mirek-fidler:
At this point: Let us wait 2 years, then switch to C++20. It is not only uppsrc, but also client apps...

klugier:
We should adopt this change sooner rather than later. The main problem is that manually switching to a newer standard pollutes compilation logs with warnings. This change won't hurt compatibility with C++17, as the updated code is perfectly valid for that standard.

I don't propose a single, massive change. My plan is to start with several packages from uppsrc and submit the fixes in a few pull requests. Ultimately, all the code within our official distribution will be updated.

For client-side code, we'll fix what's on our end. It will be the client's decision whether to fix their own code or tolerate the compilation warnings. In this scenario, we'll be clear that we aren't responsible for these warnings.

ismail-yilmaz:
> We should adopt this change sooner rather than later. The main problem is that manually switching to a newer standard pollutes compilation logs with warnings. This change won't hurt compatibility with C++17, as the updated code is perfectly valid for that standard.
> 
> I don't propose a single, massive change. My plan is to start with several packages from uppsrc and submit the fixes in a few pull requests. Ultimately, all the code within our official distribution will be updated.
> 
> For client-side code, we'll fix what's on our end. It will be the client's decision whether to fix their own code or tolerate the compilation warnings. In this scenario, we'll be clear that we aren't responsible for these warnings.

Well, for a starter, I'll see if I can get the SSH package comply with the C++20 rule without breaking anything. I think it would be trivial because all lambda calls are internal and they don't do any magic.

PR #293, opened 2025-07-19
Fix RichTxt::SetRefreshFrom logic (issue #137): move r_type assignmen…

### LINKED ISSUE #137: RichText, faulty condition check (condition is always false)
In RichTxt::SetRefreshFrom(int parti) method.

See below:
https://github.com/ultimatepp/ultimatepp/blob/2804392071c790acf8ae695a6eae1396c33106c8/uppsrc/RichText/TxtData.cpp#L107

killerdevildog:
…t after condition
Fixes #137


killerdevildog:
The SetRefreshFrom method now checks the condition before setting r_type = FROM, so the logic works as intended.

PR #297, opened 2025-09-14
Ide: show git errors when update of UppHub packages failed

klugier:
Currently there are no information when UppHub "Update all" operation ended with failures. It closes immediately and user are not aware about the problems. This PR is suppose to address this problem.

### DIFF
--- a/uppsrc/ide/UppHub.cpp
+++ b/uppsrc/ide/UppHub.cpp
@@ -540,11 +540,18 @@ void UppHubDlg::Update()
 	if(!PromptYesNo("Pull updates for all modules?"))
 		return;
 	UrepoConsole console;
+	
+	bool errors = false;
 	for(const UppHubNest& n : upv) {
 		String dir = GetHubDir() + "/" + n.name;
-		if(DirectoryExists(dir))
-			console.Git(dir, "pull --rebase");
+		if(!DirectoryExists(dir))
+			continue;
+		
+		if (console.Git(dir, "pull --rebase") != 0)
+			errors = true;
 	}
+	if (errors)
+		console.Perform();
 }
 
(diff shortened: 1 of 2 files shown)


uppsrc/ide/UppHub.cpp
```diff
@@ -540,18 +540,21 @@ void UppHubDlg::Update()
...
+
+		if(console.Git(dir, "pull --rebase") != 0)
 			errors = true;
 	}
-	if (errors)
-		console.Perform();
+	if(!errors)
+		return;
+	
+	ErrorOK("Update failed. Review the logs to diagnose and resolve the issues.");
```
ismail-yilmaz: This looks good and works as expected. However two things can be improved I guess.

1) A minor UX issue: Git console is already open and instead of opening another dialog on top of it, we can use the existing git console to report/log the final error message. 

2) This is a suggestion. Maybe reporting the number of errors would be better:

```
void UppHubDlg::Update()
{
	if(!PromptYesNo("Pull updates for all modules?"))
		return;
	UrepoConsole console;
	int errors = 0;
	for(const UppHubNest& n : upv) {
		String dir = GetHubDir() + "/" + n.name;
		if(!DirectoryExists(dir))
			continue;
		if(console.Git(dir, "pull --rebase") != 0)
			errors++;
	}
	if(!errors)
		return;
	
	String err = "Update failed (%d errors). Review the logs to diagnose and resolve the issues.";
	ErrorOK(Format(err, errors));
	console.Perform();
}
```

Other than these suggestions, It looks and works good.

klugier: Done! Thanks for feedback!

ismail-yilmaz (approved):

PR #302, opened 2025-10-11
Implement ${METHOD_DIR} recognition in umk

klugier:
Thanks to that change we will be able to bundle umk with toolchain on Windows and distribute such package. Thanks to that it will open the possibilities of build many packages including these from UppHub directly from the terminal.

Here is the sample build method for CLANG on Windows that will be handled correctly after these changes:
```
PATH = "${METHOD_DIR}\\bin\\clang\\bin;${METHOD_DIR}\\bin\\clang\\x86_64-w64-mingw32\\bin;${METHOD_DIR}\\bin\\SDL2\\lib\\x64;${METHOD_DIR}\\bin\\pgsql\\x64\\bin;${METHOD_DIR}\\bin\\mysql\\lib64";
INCLUDE = "${METHOD_DIR}\\bin\\SDL2\\include;${METHOD_DIR}\\bin\\pgsql\\x64\\include;${METHOD_DIR}\\bin\\mysql\\include;${METHOD_DIR}\\bin\\llvm";
LIB = "${METHOD_DIR}\\bin\\SDL2\\lib\\x64;${METHOD_DIR}\\bin\\pgsql\\x64\\lib;${METHOD_DIR}\\bin\\mysql\\lib64;${METHOD_DIR}\\bin\\llvm";
```

The umk Windows package will looks as follow
Zrzut ekranu 2025-10-11 184617
In above image bin folder has exactly the same content as bin directory in regular Windows U++ distribution.

### DIFF
--- a/uppsrc/umk/umake.cpp
+++ b/uppsrc/umk/umake.cpp
@@ -38,10 +38,28 @@ String Ide::GetDefaultMethod()
 	return "GCC";
 }
 
+String ReplaceMethodDir(String paths, const String& method_dir)
+{
+	constexpr const char* METHOD_DIR = "${METHOD_DIR}";
+	
+	if (paths.Find(METHOD_DIR) == -1) {
+		return paths;
+	}
+	paths.Replace(METHOD_DIR, method_dir);
+	return paths;
+}
+
 VectorMap<String, String> Ide::GetMethodVars(const String& method)
 {
 	VectorMap<String, String> map;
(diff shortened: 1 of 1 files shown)


mirek-fidler:
This is fine & merged, I just wonder:

- it should be documented somewhere
- maybe this could be in theide as well, and perhaps more of these "variables" can be added over time - in that case, it should be documented in build methods setup

PR #303, opened 2025-10-11
VirtualGui/SDL2GL: Added fullscreen mode.

ismail-yilmaz:
This PR allows the SDL2 backend of VirtualGui to use full screen mode optionally.

Please review.

### DIFF
--- a/uppsrc/VirtualGui/SDL2GL/SDL2GL.h
+++ b/uppsrc/VirtualGui/SDL2GL/SDL2GL.h
@@ -38,7 +38,7 @@ struct SDL2GUI : VirtualGui {
 	void Attach(SDL_Window *win, SDL_GLContext glcontext);
 	void Detach();
 
-	bool Create(const Rect& rect, const char *title);
+	bool Create(const Rect& rect, const char *title, bool fullscreen = false);
 	void Destroy();
 	
 	SDL2GUI();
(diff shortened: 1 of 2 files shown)


uppsrc/VirtualGui/SDL2GL/SDL2GL.h
```diff
@@ -38,7 +38,7 @@ struct SDL2GUI : VirtualGui {
 	void Attach(SDL_Window *win, SDL_GLContext glcontext);
 	void Detach();
 
-	bool Create(const Rect& rect, const char *title);
+	bool Create(const Rect& rect, const char *title, bool fullscreen = false);
```
klugier, line 41: What about turning on fullscreen mode on runtime? As I understand this will only allow to move to fullscreen mode on window creation.

ismail-yilmaz: That can be done, I suppose, but,

1) I think it would be the subject of another PR (I can also work on that separately if needed).
2) The best use case for SDL2GL I encountered so far is headless (desktopless), fullscreen mode.

I also wonder what @mirek-fidler thinks about that.

klugier: Sure, no problem on my side. Just proposing an improvement :)

PR #305, opened 2025-10-13
Screenshot changed and added

XKFS:


uppsrc/CtrlCore/CtrlDraw.cpp
```diff
@@ -414,13 +414,13 @@ void Ctrl::UpdateArea0(SystemDraw& draw, const Rect& clip, int backpaint)
 	LLOG("========== UPDATE AREA " << UPP::Name(this) << ", clip: " << clip << " ==========");
 	ExcludeDHCtrls(draw, GetRect().GetSize(), clip);
 	auto DoCtrlPaint = [&](SystemDraw& w, const Rect& clip) {
-	#ifdef PLATFORM_WIN32
+	/*#ifdef PLATFORM_WIN32
```
ismail-yilmaz: This seems to disable the TopBar on Windows 11 (which breaks VirtualGUI, which has a pending fix here #300).

klugier: Yes, this needs to be reverted to the initial version.

klugier:
You shouldn't only modify screenshot, you should also change the code to include improvements we were talking on Discord. In all examples we want window Title. It will be exactly the same to what we have in GUI tutorial. Once done, I will checkout branch and check how your changes looks like. Without these changes I marked this PR as 'do not merge'.

klugier:
@XKFS I downloaded your branch and it looks good!. However, there are things you should polish:
- formatting problem in printing example, you should use Default style for line in which image is placed and for the line above
- you modify the code in tutorials, but don't update https://github.com/ultimatepp/ultimatepp/tree/master/tutorial. Please update all six files accordingly. We just need the same GUI_APP_MAIN we have in doc.

In case of problems, please write to me on Discord :)

PR #309, opened 2025-10-16
Check git commands status in UMK UppHub logic

klugier:
In umk in UppHub code we do not check what system commands return. This PR is addressing this issue to increase UMK robustness.


klugier:
This will need rework after latest changes. Anyway, error checking is important part of experience...

PR #310, opened 2025-10-16
Core: Added size_t as a Value supported type.

klugier:
This PR add size_t as a supported type in Value class. Moreover, it added proper documentation and fix missing documentation for float type.

Motivation:
```
#include <Core/Core.h>

using namespace Upp;

CONSOLE_APP_MAIN
{
	Value a = 20;
	Value b = size_t{20}; // Not possible without change
	Value c = sizeof(int); // Not possible without change
	
	std::vector<int> vec1 = { 1, 2, 3};
	Value d = vec1.size(); // Not possible without change
}
```


mirek-fidler:
IDK, this is asking for overloading trouble...

PR #311, opened 2025-10-19
Replace apt-get with apt for Linux CI/CD build

klugier:


ped7g:
Going by my old knowledge which may be obsolete now, this is maybe technically wrong change, apt-get has stable and more powerful CLI interface while apt does not and is meant for human users.

But practically I don't think it really matters. So I'm more like curious why.. does this fix something?

klugier:
Thanks for this comment! You are right. I reverted to apt-get https://github.com/ultimatepp/ultimatepp/commit/46fdd0e966cb71b2b5a32eb6de5849f97f3fbd3f.

PR #313, opened 2025-10-21
Screenshot doc default style and tutorial draw rewrite

XKFS:
PR #305 but fixed the screenshot error and draw rewritten


klugier:
This PR is a mess, 64 files has been changed. Please fix in #305. I only asked for updating 6 cpp files. This should be doable.

PR #315, opened 2025-10-22
Eliminate crash in Ctrl::GetOwner() when short live popup is being displayed

klugier:
Latest version of U++ on Gtk crashes when hover on element that needs to be expanded via tool-tip to display full content. This PR is fixing this problem.


mirek-fidler:
Good catch, but little too many lines. How about

return top && IsOpen() ? top->owner : nullptr;

klugier:
Changed as suggested.

mirek-fidler:
Thank you!

PR #316, opened 2025-10-26
Ide: added package count for UppHub dialog

klugier:
This is similar changes to what we have in PackageSelector, but this time it is for UppHub dialog. It just displays how many elements are in array.

UppHubPackageCount


ismail-yilmaz (approved):
Looks good. Small touches are nice.

klugier:
BTW, Should we rename Name column to Package? Please let me know what do you think @mirek-fidler and @ismail-yilmaz? I think name is not the best option here, but what will be batter...

ismail-yilmaz:
> BTW, Should we rename Name column to Package? Please let me know what do you think @mirek-fidler and @ismail-yilmaz? I think name is not the best option here, but what will be batter...

Yes, IMO,  "Package" would describe them better.

mirek-fidler:
> BTW, Should we rename Name column to Package? Please let me know what do you think @mirek-fidler and @ismail-yilmaz? I think name is not the best option here, but what will be batter...

I am not entirely sure. Yes, usually it contains one important package, but technically each entry is a sort of "nest" that can contain multiple packages.

ismail-yilmaz:
> > BTW, Should we rename Name column to Package? Please let me know what do you think @mirek-fidler and @ismail-yilmaz? I think name is not the best option here, but what will be batter...
> 
> I am not entirely sure. Yes, usually it contains one important package, but technically each entry is a sort of "nest" that can contain multiple packages.

Yes, but IMO that's a very "uppish" technical standpoint. :) Granted, they *are* upp code, but from user standpoint they are probably better treated as packages, which is consistent with other package managers, hence less confusion (for users).

mirek-fidler:
OK, please rename to "Package".

klugier:
@mirek-fidler done!

uppsrc/ide/UppHub.cpp
```diff
@@ -92,6 +92,8 @@ void UppHubSettingsDlg::RefreshCtrls()
 }
 
 struct UppHubDlg : WithUppHubLayout<TopWindow> {
+	static constexpr auto LIST_KEY = "PACKAGE";
```
mirek-fidler, line 95: Nope.

uppsrc/ide/UppHub.cpp
```diff
@@ -144,8 +146,8 @@ UppHubDlg::UppHubDlg()
 	parent.Add(list.SizePos());
 	parent.AddFrame(splitter.Right(info, 500));
 	
-	list.AddKey("NAME");
```
mirek-fidler, line 147: No reason to change this.

mirek-fidler (commented):
We might agree to disagree but by this point it should be clear that I refuse to add noise lines into the code.... (I have accepted one such noisy affair in CommandLineHandler in ide and to this day I regret it - will probably remove it ASAP).

klugier:
Thanks for flagging this, Mirek. This is exactly why code reviews are crucial. I wasn't aware you preferred a different approach for handling that predefined string declaration used across multiple areas.

Historically when I worked with JNI, I found that using a single, shared string declaration was the best practice to maintain consistency and prevent human error when the value needed to be reused across the codebase.

I'll keep this design preference in mind for future work. Thanks!

mirek-fidler:
> Thanks for flagging this, Mirek. This is exactly why code reviews are crucial. I wasn't aware you preferred a different approach for handling that predefined string declaration used across multiple areas.

Let me explain a bit here: I think that the key here is "multiple areas". In this case, the string was used in single file some 10 lines apart - that can hardly be defined as "multiple areas" :)

If it was important part of interface that is really used **across the codebase** it might be another story entirely. Here it is really just implementation detail. So please use the common sense. E.g. I do not have problem with

class AndroidManifest {
public:
	static constexpr const char* FILE_NAME = "AndroidManifest.xml";

In general, I believe that minimizing the length of sources is quality of its own. Adding constructs that do not improve anything like unnecessary class factories, base classes, replacing 10 lines of codes with 100 lines just to please some design pattern gods: Nope :)

I want U++ lean and mean. Actually, there is hardly another option, we do not have resource to maintain large codebase anyway :)

(P.S.: https://github.com/ultimatepp/ultimatepp/commit/29cf8728ecae697e01791e131f9ea65546d60fc0)

PR #318, opened 2025-11-02
Removed using namespace Upp from SIMD_NEON.h

klugier:
Obvious problem since this file is currently under Upp namespace. Also, it polutes all files that include Core/Core.h.


mirek-fidler:
Thanks, this was pretty stupid....

klugier:
@mirek-fidler this change has not been merged!

PR #324, opened 2025-11-23
plugin/ppm: PPMRaster, parser fix

ismail-yilmaz:
This PR addresses a parsing issue in the original `plugin/ppm` package, aligning its behavior with the [official PPM specification](https://netpbm.sourceforge.net/doc/ppm.html).

### Included Fixes

* Properly skips all whitespace.
* Correctly ignores all comment lines.

Please review

### DIFF
--- a/uppsrc/plugin/ppm/ppm.cpp
+++ b/uppsrc/plugin/ppm/ppm.cpp
@@ -21,10 +21,9 @@ bool PPMRaster::Create()
 		for(;;) {
 			while(IsSpace(stream.Peek()))
 				(void) stream.Get();
-			if(stream.Peek() == '#')
-				(void) stream.GetLine();
-			else
+			if(stream.Peek() != '#')
 				break;
+			[[maybe_unused]] String s = stream.GetLine();
 		}
 	};
 	
@@ -48,19 +47,19 @@ bool PPMRaster::Create()
 	}
 	
 	SkipWhitespaces();
-	if(size.cx = ReadInt(); size.cx <= 0 || size.cx > 99999) {
+	if(size.cx = ReadInt(); size.cx <= 0 || size.cx >= std::numeric_limits<word>::max()) {
 		SetError();
 		return false;
 	}
 	
 	SkipWhitespaces();
-	if(size.cy = ReadInt(); size.cy <= 0 || size.cy > 99999) {
+	if(size.cy = ReadInt(); size.cy <= 0 || size.cy >= std::numeric_limits<word>::max()) {
 		SetError();
 		return false;
 	}
 	
 	SkipWhitespaces();
-	if(int maxval = ReadInt(); maxval > 0 && maxval < 65536) {
+	if(int maxval = ReadInt(); maxval > 0 && maxval < std::numeric_limits<word>::max()) {
 		SkipWhitespaces();
 		is16 = maxval > 255;
 		pixel_pos = stream.GetPos();
--- a/uppsrc/plugin/ppm/ppm.cpp
+++ b/uppsrc/plugin/ppm/ppm.cpp
@@ -16,39 +16,58 @@ bool PPMRaster::Create()
 		SetError();
 		return false;
 	}
-
+	
+	auto SkipWhitespaces = [&] {
+		for(;;) {
+			while(IsSpace(stream.Peek()))
+				(void) stream.Get();
+			if(stream.Peek() == '#')
+				(void) stream.GetLine();
+			else
+				break;
+		}
+	};
+	
 	auto ReadInt = [&] {
 		int n = 0, c;
-		while((c = stream.Get()) >= '0' && c <= '9')
+		while((c = stream.Get()) >= '0' && c <= '9') {
+			// Overflow check: ensure n * 10 + digit doesn't overflow
+			if(n > (INT_MAX - 9) / 10) {
+				SetError();
+				return -1;
+			}
 			n = n * 10 + (c - '0');
+		}
 		return n;
 	};
-
-	auto SkipWhitespaces = [&] {
-		while(IsSpace(stream.Peek()))
-			(void) stream.Get();
-		while(stream.Peek() == '#') // PPM can contain comment lines
-			(void) stream.GetLine();
-	};
-
+	
 	// PPM P6 format
-	if(stream.Get() == 'P' && stream.Get() == '6') {
-		SkipWhitespaces();
-		if(size.cx = ReadInt(); size.cx <= 0 || size.cx > 99999) {
-			return false;
-		}
-		SkipWhitespaces();
-		if(size.cy = ReadInt(); size.cy <= 0 || size.cy > 99999) {
-			return false;
-		}
-		SkipWhitespaces();
-		if(int maxval = ReadInt(); maxval > 0 && maxval < 65536) {
-			SkipWhitespaces();
-			is16 = maxval > 255;
-			pixel_pos = stream.GetPos();
-			return !stream.IsEof();
-		}
+	if(stream.Get() != 'P' || stream.Get() != '6') {
+		SetError();
+		return false;
 	}
+	
+	SkipWhitespaces();
+	if(size.cx = ReadInt(); size.cx <= 0 || size.cx > 99999) {
+		SetError();
+		return false;
+	}
+	
+	SkipWhitespaces();
+	if(size.cy = ReadInt(); size.cy <= 0 || size.cy > 99999) {
+		SetError();
+		return false;
+	}
+	
+	SkipWhitespaces();
+	if(int maxval = ReadInt(); maxval > 0 && maxval < 65536) {
+		SkipWhitespaces();
+		is16 = maxval > 255;
+		pixel_pos = stream.GetPos();
+		return !stream.IsEof();
+	}
+	
+	SetError();
 	return false;
 }
 
--- a/uppsrc/plugin/ppm/ppm.cpp
+++ b/uppsrc/plugin/ppm/ppm.cpp
@@ -17,27 +17,39 @@ bool PPMRaster::Create()
 		return false;
 	}
 
-	try {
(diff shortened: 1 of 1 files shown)


uppsrc/plugin/ppm/ppm.cpp
```diff
@@ -16,28 +16,59 @@ bool PPMRaster::Create()
...
-		if(maxval <= 0 || maxval > 65535)
-			return false;
-		is16 = maxval > 255;
-		pixel_pos = stream.GetPos();
-		return true;
+	
+	auto SkipWhitespaces = [&] {
+		for(;;) {
+			while(IsSpace(stream.Peek()))
+				(void) stream.Get();
```
klugier, line 23: [[maybe_unused]] attribute is cleaner solution. https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html

uppsrc/plugin/ppm/ppm.cpp
```diff
@@ -16,28 +16,59 @@ bool PPMRaster::Create()
...
+			if(stream.Peek() == '#')
+				(void) stream.GetLine();
+			else
+				break;
```
klugier: else is not needed
```
if(stream.Peek() != '#')
    break;
[[maybe_unused]] stream.GetLine();
```

uppsrc/plugin/ppm/ppm.cpp
```diff
@@ -16,28 +16,59 @@ bool PPMRaster::Create()
...
+	};
+	
+	// PPM P6 format
+	if(stream.Get() != 'P' || stream.Get() != '6') {
+		SetError();
+		return false;
+	}
+	
+	SkipWhitespaces();
+	if(size.cx = ReadInt(); size.cx <= 0 || size.cx > 99999) {
```
klugier: What is the origin of 99999? Consider defining it as a constant. This limit has been used in two places.

uppsrc/plugin/ppm/ppm.cpp
```diff
@@ -16,28 +16,59 @@ bool PPMRaster::Create()
...
-	catch(CParser::Error) {
+	
+	SkipWhitespaces();
+	if(size.cy = ReadInt(); size.cy <= 0 || size.cy > 99999) {
+		SetError();
 		return false;
 	}
+	
+	SkipWhitespaces();
+	if(int maxval = ReadInt(); maxval > 0 && maxval < 65536) {
```
klugier: std::limits can be used in this place to represent 65536 as unsigned short int limit.

klugier:
I added some comments. We should also consider adding some auto tests for this plugin.

uppsrc/plugin/ppm/ppm.cpp
ismail-yilmaz: Done.

uppsrc/plugin/ppm/ppm.cpp
ismail-yilmaz: Done.

uppsrc/plugin/ppm/ppm.cpp
ismail-yilmaz: Apparenty, it is arbitrary. I removed them and used `std::numeric_limits<word>::max()` for all.

ismail-yilmaz:
> I added some comments. We should also consider adding some auto tests for this plugin.

Thanks, I applied the changes.
I can also add autotest for the plugin, but it should be another PR.

mirek-fidler:
PPM specifies 65536, not C++ short int limit, which can be anything >=
65535.

You are not only adding noise, you are adding a bug.

uppsrc/plugin/ppm/ppm.cpp
mirek-fidler: It is interesing that you have put [[maybe_unused]] only on one of these...

However, I am not sure what is the purpose putting (void) there anyway. I am still operating on general agreement that most return values can be safely ignored, unless they are [[nodiscard]], which U++ is not using yet and if was, it would not be using these for Get / GetLine as those are litterally meant for skipping. Has anything changed?

mirek-fidler:
So, to make it clear, the problem this fixes compared to current code is that

- placement of values on lines can be arbitrary
- line starting with # (after whitespaces) should be ignored

Correct?

I feel a bit uneasy about reimplementing GetInt, it is code duplication and overflow detection is not all that simple.

All I see is that we need to read 3 integers from the file, while skipping lines with #. I guess that can be done with  simpler code, using CParser, right?

(If you happen to have good ppm with all problems we are fixing, I can do it...)

ismail-yilmaz:
> So, to make it clear, the problem this fixes compared to current code is that
> 
>     * placement of values on lines can be arbitrary
> 
>     * line starting with # (after whitespaces) should be ignored
> 
> 
> Correct?

Yes,

The current `ppm` package assumes that header fields (magic, `cx`, `cy`, `maxval`) are separated only by line feeds, while the PPM specification allows any whitespace. Only the final header line must end with an `LF`.

It also fails on PPM files produced by recent GIMP versions, which insert a comment line into the header.

PPM may be old, but it's still used, and I’ve run into this issue multiple times, which led me to identify the problem.

ismail-yilmaz:
@mirek-fidler 

> I feel a bit uneasy about reimplementing GetInt, it is code duplication and overflow detection is not all that simple.


I'm just curious, aside from the code duplication, what's wrong with the overflow detection in this specific case? PPM expects only positive integers, and the proposed code is already checking for its range. both inside ReadInt and out.

ismail-yilmaz:
The "noisy" version can also be refactored like this (I understand you prefer CParser):

```cpp
bool PPMRaster::Create()
{
	Stream& stream = GetStream();
	size = Size(0, 0);
	if(!stream.IsOpen()) {
		SetError();
		return false;
	}
	
	auto SkipWhitespaces = [&] {
		for(;;) {
			while(IsSpace(stream.Peek()))
				(void) stream.Get();
			if(stream.Peek() != '#')
				break;
			[[maybe_unused]] String s = stream.GetLine();
		}
	};
	
	auto ReadInt = [&] {
	    SkipWhitespaces();
	    int n = 0, c;
	    while((c = stream.Get()) >= '0' && c <= '9') {
	        if(n > (65535 - 9) / 10)
	            throw Exc();
	        n = n * 10 + (c - '0');
	    }
	    if(n < 1)
	        throw Exc();
	    return n;
	};
	
	try {
		if(stream.Get(2) != "P6")
			throw Exc();
		size.cx = ReadInt();
		size.cy = ReadInt();
		is16    = ReadInt() > 255;
		SkipWhitespaces();
		pixel_pos = stream.GetPos();
		return !stream.IsEof();
	}
	catch(const Exc&) {
		SetError();
		return false;
	}
}
```

mirek-fidler:
First of all, I sort of appologize for nitpicking... Anyway, I decided to be a bit more harsh about "noise".

Anyway, my reviews probably got lost somewhere, so:

	auto SkipWhitespaces = [&] {
		for(;;) {
			while(IsSpace(stream.Peek()))
				(void) stream.Get();
			if(stream.Peek() != '#')
				break;
			[[maybe_unused]] String s = stream.GetLine();
		}
	};

I wonder why we have (void) in one instance, then [[maybe_unused]] String s in other. GetLine and Get are meant to be able to "skip" with return value ignored and do not have [[nodiscard]], so what is the point for either? Did compiler complain?

Also, I think it is good to have sanity checks (99999) (which I see are replaced be overflow tests), but maybe the code should be even more fortified (e.g. check that there is enough data in the file, or use Huge to load data...) - of course, current code does not do that either...

In any case, you got me interested, can I have offending .ppm please?

ismail-yilmaz:
@mirek-fidler 

> First of all, I sort of appologize for nitpicking... Anyway, I decided to be a bit more harsh about "noise".

No worries, I don't mind nitpicking PRs.

I haven't seen your review (notification seem to have gone unnoticed), apologize for that.


> I wonder why we have (void) in one instance, then [[maybe_unused]] String s in other. GetLine and Get are meant to be able to "skip" with return value ignored and do not have [[nodiscard]], so what is the point for either? Did compiler complain?

I added `void` and later `[[maybe_unused]]` (as per Klugier's suggestion) to warn the maintainer(s), not the compiler, that not using the result is deliberate. (An old habit of mine. Probably I should drop that...) 


> In any case, you got me interested, can I have offending .ppm please?

Attached the [gimp](https://github.com/user-attachments/files/23712097/upp.-gimp.zip) version (upp logo), which always include a comment line.




There is  (several, actually) also another version which contains "P6 cx cy maxval" **in a single line**, but I don't have access to them right now, I'll send one ASAP.

mirek-fidler:
Mostly for fun, to serve mine strong NIH syndrome...

```
bool PPMRaster::Create()
{
	Stream& stream = GetStream();
	size = Size(0, 0);
	if(!stream.IsOpen()) {
		SetError();
		return false;
	}

	try {
		if(stream.Get(2) != "P6")
			return false;
		int    ii = 0;
		int    num[3];
		String line;
		while(ii < 3) {
			CParser p(line);
			while(p.IsEof() || p.Char('#')) {
				if(stream.IsEof())
					return false;
				line = stream.GetLine();
				p.Set(line);
			}
			int n = p.ReadInt();
			if(n <= 0 || n > 65535)
				return false;
			num[ii++] = n;
		}
		size.cx = num[0];
		size.cy = num[1];
		is16 = num[2] > 255;
		pixel_pos = stream.GetPos();
		return true;
	}
	catch(CParser::Error) {
		return false;
	}
}
```

BTW both yours and this version are in violation in PPM specs:

8. A **single whitespace** character (usually a newline). 

But mine is probably less problematic, yours can IMO run into image data if it starts with whitspace (not unlikely)

ismail-yilmaz:
> But mine is probably less problematic, yours can IMO run into image data if it starts with whitspace (not unlikely)

This with a whitespace check in the end would be just fine. Thanks.

PR #325, opened 2025-11-30
Core: UnixSocket (AF_UNIX) implementation

ismail-yilmaz:
# **UnixSocket (AF_UNIX) implementation**

## **Overview**
This PR adds `UnixSocket` class for Unix domain socket (AF_UNIX) support, complementing existing `TcpSocket`. 

## **Features**
- Filesystem and abstract sockets (Linux)
- Peer process identification via PID (Linux/macOS/FreeBSD)
- Full TcpSocket I/O compatibility
- Cross-platform POSIX support
 
## **Benefits**
- Local IPC performance over TCP
- Process authentication via peer PID
- Clean API with full TcpSocket feature set
- No breaking changes to existing code

### **Modified Files**
- `Core/Inet.h` - Added `friend class UnixSocket` to TcpSocket
- `Core/Core.h` - Added header include line

### **New Files**
- `Core/UnixSocket.h`- UnixSocket class declaration
- `Core/UnixSocket.cpp` - Implementation

### **New Examples**
- `reference/UnixSocketServer` - Demo server application (replicates SocketServer example)
- `reference/UnixSocketClient` - Demo client application (replicates SocketClient example)

### **Unit tests**
-  `autotest/UnixSocket` - Tests server/client (local IPC) communication (connect, listen, accept)

Please review.


mirek-fidler:
mutable int     peerpid;

In general, const method, lazy evaluation and 'mutable' is trouble for MT. And I think it might be the case here...

GetPeerPid is const and thus in general it should be possible to be invoked by 2 threads without locking. If not, it needs a lot of explanation in help for unexpected behaviour, but I guess it should be possible to do it right....

ismail-yilmaz:
> mutable int peerpid;
> 
> In general, const method, lazy evaluation and 'mutable' is trouble for MT. And I think it might be the case here...
> 
> GetPeerPid is const and thus in general it should be possible to be invoked by 2 threads without locking. If not, it needs a lot of explanation in help for unexpected behaviour, but I guess it should be possible to do it right....

@mirek-fidler 

Yeah, looking back again, it is unnecessarily complex anyway. I refactored the `GetPeerPid` method and completely removed the unnecessary `peerpid` variable. It now directly calls the `getsockopt` function. Docs also state that `GetPeerPid` in NB mode should be called after the connection is actually established, otherwise it'll return -1.

Please check.

ismail-yilmaz:
@mirek-fidler 

Anything else should I need to do with this? With removeing the MT sensitive stuff situation should be much better now.

mirek-fidler:
Late in the night I have got this little worry:

Are there algorithms that could be shared between TcpSocket and UnixSocket? (e.g. GUI widget loading raw data from socket which should work both with TcpSocket and UnixSocket), If yes, then maybe we should reconsider separate class and get back to single interface?

(Sorry that it only came to mind late in the process)

ismail-yilmaz:
> Late in the night I have got this little worry:
> 
> Are there algorithms that could be shared between TcpSocket and UnixSocket? (e.g. GUI widget loading raw data from socket which should work both with TcpSocket and UnixSocket), If yes, then maybe we should reconsider separate class and get back to single interface?
> 
> (Sorry that it only came to mind late in the process)

Fair point. My initial reason for introducing UnixSocket as a separate class was to avoid complicating TcpSocket with Unix-specific logic. However, if we want both socket types to work transparently, or allow polymorphism, and since their I/O logic is virtually identical, then a single interface is possible. So we can:

a) Add Unix socket support as a mode in the existing class, using a simple switch or a small set of boolean flags (for filesystem or abstract socket types).

b) Introduce a base class, ASocket, with TcpSocket and UnixSocket as subclasses. This is not difficult. We can make five or six methods protected and both subclasses will work cleanly. Alternatively, we can use an intentional antipattern and avoid protected methods entirely by making both subclasses friends of the base class. It is unconventional, but it would work. The only issue I can foresee is method chaining. We can either duplicate the chainable methods in both subclasses or make the base class a template that returns T. The template approach is cleaner, but also more verbose.

I can prototype both options, no big deal. Which direction would you prefer?

mirek-fidler:
Lookig at the code, it seems to me that all that is needed is to add

ConnectFilesystem
ConnectAbstract
ListenFilesystem
ListenAbstract

methods to TcpSocket (and then rename TcpSocket to something and typedef as obsolete). Accept seems to be the same.

I do not think that TcpSocket even does need to remember in what mode it is open.

ismail-yilmaz:
> Lookig at the code, it seems to me that all that is needed is to add
> 
> ConnectFilesystem ConnectAbstract ListenFilesystem ListenAbstract
> 
> methods to TcpSocket (and then rename TcpSocket to something and typedef as obsolete). Accept seems to be the same.
> 
> I do not think that TcpSocket even does need to remember in what mode it is open.

OK, but do we really need to rename it at this point? I mean, we can just state their operation mode in the documentation and make these methods static assert on non posix systems if ınvoked. Any call to connect/listen variants will clear the socket and set it into desired mode anyway. It'll be even simpler.

mirek-fidler:
> OK, but do we really need to rename it at this point?

Maybe there is a bit of confusion of what is renamed: TcpSocket. Only because it is now just Socket that can be open in TCP or filesystem or abstract mode, so the name does not fit anymore. That said, not really necessarry indeed, I can live with TcpSocket with UnicSocket features...

> make these methods static assert on non posix systems if ınvoked.

Even better, make them disappear with #ifdef PLATFORM_POSIX ...

ismail-yilmaz:
> > OK, but do we really need to rename it at this point?
> 
> Maybe there is a bit of confusion of what is renamed: TcpSocket. Only because it is now just Socket that can be open in TCP or filesystem or abstract mode, so the name does not fit anymore. That said, not really necessarry indeed, I can live with TcpSocket with UnicSocket features...
> 
> > make these methods static assert on non posix systems if ınvoked.
> 
> Even better, make them disappear with #ifdef PLATFORM_POSIX ...


Ah ok then. I'll,

- Rename the `TcpSocket` as `Socket`
- Implement `ConnectFileSystem`, `ConnectAbstract`, `ListenFileSystem`, `ListenAbstract` and `GetPeerPid` methods for POSIX environments only.
- Alias `Socket` as `TcpSocket` AND `UnixSocket` (posix only)
- Add reference examples for UnixSocket (same as socket server/client examples)
- Add autotest for UnixSocket on POSIX.
- Update the socket docs accordingly.

right?

mirek-fidler:
"Socket" - unless there is a nameclash (hopefully there is not)
"UnixSocket" - TcpSocket is meant to be obsolete name. Do we want to add new "UnixSocket" that is already "obsolete"?

ismail-yilmaz:
> "Socket" - unless there is a nameclash (hopefully there is not) "UnixSocket" - TcpSocket is meant to be obsolete name. Do we want to add new "UnixSocket" that is already "obsolete"?

Ok, we can ditch `UnixSocket` alias too. It's just a functioning mode. API docs will be sufficient I guess.

ismail-yilmaz:
Ok, Here's what I did:

- Renamed `TcpSocket` as `Socket`
- Set `TcpSocket` as an alias for `Socket` for backward compat.
- Changed `SocketWaitEvent` Add() function signature to `Socket`
- Added required functions with conditional compilation.
- Updated the API docs.
- Updated reference examples and autotest (for unix socket part)
- removed UnixSocket.h and UnixSocket.cpp completely.

It works well as-is now. However, if anything is missing or needs more work, let me know.

Please check.

ismail-yilmaz:
A note: I made the `ConnectAbstract` and `ListenAbstract` methods available on all Posix systems, although the abstract socket concept is only available on Linux. As it is stated in their API docs, these functions will return false and set the socket into error state in other Posix-compliant OSes. Namely they are not hidden behind PLATFORM_LINUX guards.

mirek-fidler:
Excellent, thank you.

PR #326, opened 2025-12-04
Fix Circle CI on macOS

klugier:


mirek-fidler:
Is it all? :)

klugier:
Yes, We can merge.

PR #327, opened 2025-12-05
CtrlLib: Fixed disabled style of OptionButton

ortman:


ortman:
When the button is pressed (option=true) and disabled, the style remains CTRL_PRESSED, but CTRL_DISABLED is needed.

mirek-fidler:
Thanks.

PR #328, opened 2025-12-13
Core/Socket: Unix domain socket (AF_UNIX) support for Windows

ismail-yilmaz:
This PR adds support for **AF_UNIX (Unix domain sockets) on Windows** using native Winsock functionality available in Windows 10 (1803+) and newer.

(Tom informed me about the availability of AF_UNIX on Windows, and here we are.)

### Notes

* AF_UNIX is enabled only when available at compile time (`#ifdef AF_UNIX`)
* Uses `<afunix.h>` on Windows for `sockaddr_un`
* Linux-only features (abstract sockets, peer credentials) are explicitly rejected on other platforms
* Exposes AF_UNIX methods of Socket to all platforms but ensures graceful failure on unsupported systems.

### Limitations on Windows

* Path-based UNIX sockets only
* No abstract socket namespace
* No peer PID / credential support

### Other
* Examples and autotest are updated.
* POSIX behavior is unchanged.

Please check.

### DIFF
--- a/autotest/UnixSocket/UnixSocket.cpp
+++ b/autotest/UnixSocket/UnixSocket.cpp
@@ -2,50 +2,63 @@
 
 using namespace Upp;
 
+String GetSocketPath()
+{
+	String temp;
+#ifdef PLATFORM_WIN32
+	temp = GetEnv("TEMP");
+#else
+	temp = GetTempPath();
+#endif
+	return AppendFileName(temp, "upp-unixsocket.socket");
+}
+
 CONSOLE_APP_MAIN
 {
-#ifdef PLATFORM_POSIX
-
 	StdLogSetup(LOG_COUT|LOG_FILE);
 
-	String path = Format("/tmp/upp-unixsocket-test-%d", getpid());
-	
-	Socket server, client;
-	
-	// Test server listen
-	if(!server.ListenFileSystem(path)) {
-		LOG("Server listen failed: " << server.GetErrorDesc());
-		Exit(1);
-	}
-	
-	// Test client connect
-	if(!client.ConnectFileSystem(path)) {
-		LOG("Client connect failed: " << client.GetErrorDesc());
-		Exit(1);
-	}
-	
-	// Test data exchange
-	String test_data = "Hello, world!";
-	client.Put(test_data + "\n");
-	
-	Socket accepted;
-	if(!accepted.Accept(server)) {
-		LOG("Accept failed: " << accepted.GetErrorDesc());
-		Exit(1);
-	}
-	
-	String received = accepted.GetLine();
-	DUMP(received);
-	
-	ASSERT(received == test_data);
-	
-	// Test peer PID (on supported platforms)
-	int pid = accepted.GetPeerPid();
-	DUMP(pid);
-	if(pid != -1)
-		ASSERT(pid == getpid()); // Should be our own process in this test
+	String path = GetSocketPath();
 
-	LOG("=========== OK");
+	try {
+		Socket server, client;
 
-#endif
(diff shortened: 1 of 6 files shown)


User0755:
Hi Ismail,

The ultimatepp.org site seems to be down again. (Wonder what's wrong with the server these days.)

Anyway, I managed to test the UnixSockets on Windows 11. It compiles without trouble using both CLANG and MSBT tools. The updated UnixSocketServer.cpp and UnixSocketClient.cpp can also communicate with each other regardless of the compiler used. However, the socket path derived with GetTempDirectory() does not work on Windows. It returns a unique temporary path every time when called, so the Client and Server cannot find each other. When using const String& path = "C:\\Temp\\upp-unixsocket.sock"; instead, it works fine.

Thanks and best regards,

Tom

ismail-yilmaz:
> Hi Ismail,
> 
> The ultimatepp.org site seems to be down again. (Wonder what's wrong with the server these days.)
> 
> Anyway, I managed to test the UnixSockets on Windows 11. It compiles without trouble using both CLANG and MSBT tools. The updated UnixSocketServer.cpp and UnixSocketClient.cpp can also communicate with each other regardless of the compiler used. However, the socket path derived with GetTempDirectory() does not work on Windows. It returns a unique temporary path every time when called, so the Client and Server cannot find each other. When using const String& path = "C:\Temp\upp-unixsocket.sock"; instead, it works fine.
> 
> Thanks and best regards,
> 
> Tom

Hi Tom,
I've changed the path. Should work now (it works on my Windows machines).
Please check.

Best regards,
İsmail

User0755:
Hi Ismail,

Thanks! Yes it worked here too. I'm still a bit worried about the C:\Windows\temp\ since it might not be accessible to all users.

Here's what copilot.microsoft.com thinks about the AF_UNIX socket path selection:

AF_UNIX sockets on Windows are a bit special compared to Linux/Unix, because Windows doesn’t have a traditional /tmp directory. Instead, you need to choose a location that:

    ✅ Exists for all users

    ✅ Is writable by normal (non‑admin) accounts

    ✅ Has a predictable path

    ✅ Doesn’t require elevated privileges

Recommended Choices

    %TEMP% or %TMP% (per‑user temp directory)

        Example: C:\Users\<username>\AppData\Local\Temp

        Pros: Always writable by the user, no admin rights needed.

        Cons: Only accessible to that user — not shared across accounts.

    C:\Windows\Temp (system temp directory)

        Pros: Accessible to multiple users.

        Cons: Permissions can vary; sometimes restricted. Not ideal for cross‑user communication unless you carefully manage ACLs.

    A dedicated application‑specific directory under ProgramData

        Example: C:\ProgramData\MyApp\sockets

        Pros: Shared across users, predictable, and you can set ACLs to allow normal users.

        Cons: Requires setup (creating the directory and setting permissions).

--

Best regards,

Tom

ismail-yilmaz:
The actual issue here is that `GetTempPath()` (the U++ one) returns a path relative to the executable/U++ folder on Windows. As a result, each application instance ends up with a different path. Or at least, that's what I've observed in logs. Trivial to fix though. we'll get it from env. :) 

Until I clear this issue, marking this PR as a draft. :)

Thanks for testing, and your patience.

Best regards,
İsmail

ismail-yilmaz:
@mirek-fidler 

I fixed the examples, autotest and added a note about the lifetime of the socket path/file to `ListenFileSystem` method API doc.

AF_UNIX for Windows Socket patch is now ready for review.

User0755:
Hi Ismail,

Yes, that seems a solid choice.

Best regards,

Tom

PR #330, opened 2025-12-17
Draw Tutorial screenshot and code updated

XKFS:
I don't know how to add this to pr-305 so had to create one


mirek-fidler:
It is improvement, but I guess we will need to redo the whole thing soon anyway.

Thanks.

PR #331, opened 2025-12-20
Core: NoCopy modernization and documentation.

klugier:
I did modernization and replace private with delete methods. This should give better compiler output when some class consumer hit the problem.


ismail-yilmaz (approved):
Lgtm

PR #332, opened 2025-12-25
ide: Support for creating git and hpp files

ortman:
Added exceptions for creating .gitignore and .gitmodules files from IDE. And a template for new HPP files.

### DIFF
--- a/uppsrc/ide/NewPackageFile.cpp
+++ b/uppsrc/ide/NewPackageFile.cpp
@@ -88,7 +88,11 @@ String NewPackageFileWindow::GetError()
 	if(FileExists(p))
 		return String().Cat() << "File&[* \1" << p << "\1]&already exists!";
 	if(*n == '.') {
-		Index<String> exceptions = {".clang-format"};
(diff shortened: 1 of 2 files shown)


mirek-fidler:
I am a bit uneasy about .hpp files. It is mostly u++ idiosyncrasy, but we are usually using .hpp as extension for files that are included but actually contain implementation. I do not think there is a single U++ .hpp file with include guards in the codebase...

ortman:
Some projects have many HPP files, and since they are included via #include, re-including them in other project files will result in a redefinition error.
Therefore, protection in HPP files is a good idea (similar to C-headers).

mirek-fidler:
I do not dispute that, it is just that .hpp has different meaning in U++.

That said, one can always delete those guards, right?

ortman:
Of course, this is just a template for file creation; no one is required to use it..
If HPP uses a header file containing the implementation, you can remove #ifndef-#define-#endif or leave it in place, as it won't break compilation. Each programmer will decide for themselves in their own project.

PR #333, opened 2025-12-25
Arrayctrl link fix

XKFS:


klugier:
This is partially done. We need to fix ALL links in the article. Some of the links not fixed in this PR are: 
- Order
- Image
- String
- IdInfo
- Id
- Column
- Vector
- HeaderCtrl::Column
- ... (you just need to find the rest of them)

I also suggesting disable "Show spelling error" option from the first row of icons in Topic editor since it might hide links.

klugier:
Code review:
- ArrayCtrl& OddRowColor(Color paper = SColorInfo, Color ink = SColorText) (The first color is still broken)
- ArrayCtrl& AcceptEdits(bool b = true) (ArrayCtrl in return should be link)
- Id is broken and in general in IdInfo& AddKey(const Id& id) it refers to wrong symbol in documentation.
- Column still seems to be broken. However, I noticed that in some places you are doing it good. So, in the end you need to double check and fix all places.
- Ctrl doesn't have a link at least in IdInfo& AddCtrl(Ctrl& ctrl). In other places seems to be no link. This is easy to fix since we have documenation for this class.
- HeaderCtrl and HeaderCtrl::Column - on latest master you can link to this element properrly
- No link to Stream
- No link to Display
- No link to Rect (You can look where I link it in HeaderCtrl)
- No link to Point
- No link to Value
- No link to ValueArray
- No link to ValueMap
- No link to Nuller (Thanks to my recent changes there is a possibility to create the link to it)
- Gate2 not documented should be in one of article in Core, I assume callbacks or something similar.
- Not link to Bar
- Callback should be replaced with Event and the link. You can look how I did it in latest headerCtrl
- No link to Function
- No link to One

In general you did a great progress, but still there is a need to figure out some things and improve links.

I also adding the link to my recent changes in HeaderCtrl https://github.com/ultimatepp/ultimatepp/commit/0044e5ed485372d1cd46a7f46e549820d7c8f8ed. It will give you better visibility on things I fixed and you might find it as an inspiration to some of your changes.

PR #335, opened 2025-12-28
Arrayctrl tpp link fix

XKFS:


klugier (approved):

klugier:
I will merge. However, still not all symbols are being covered and Id is being link wrongly in some cases (not all).

PR #337, opened 2026-01-01
CtrlCore: add support for mouse back and forward keys on macOS.

klugier:


mirek-fidler:
Brevity please...

(No need for separate function, no need to introduce constants when you comment that 4 and 5 are standards etc... This can be collapsed into 3-5 lines)

klugier:
Separate function is needed as DispatchKey is private and it can not be directly called by CocoView implementation. Please keep in mind that this is the reason that most of the function in this interface implementation works like that. I simplified the logic as requested. Thanks for keeping an eye on that.

klugier:
Thanks for rework! Didn't know that decode exists.

PR #339, opened 2026-01-03
CtrlCore: Cocoa always use bundled icon

klugier:
This PR introduces SetAlwaysUseBundledIcon method for Cocoa. It supposed to globaly disable SyncAppIcon operation. Thanks to that the app that uses this function is more consistent with the macos feeling. Here is the samples:

1. Before
Zrzut ekranu 2026-01-3 o 02 05 05
Zrzut ekranu 2026-01-3 o 02 06 31

2. After
Zrzut ekranu 2026-01-3 o 02 07 37
Zrzut ekranu 2026-01-3 o 02 08 05

For 'After' the app icon in the dock during the runtime scales to the os settings. Without this option it does not.

Please feel free to rename the funcionality to whatever you want. I just tried to use your notation, but maybe we can name it better.

### DIFF
--- a/uppsrc/CtrlCore/CocoCtrl.h
+++ b/uppsrc/CtrlCore/CocoCtrl.h
@@ -2,8 +2,10 @@ private:
 	friend struct MMCtrl;
 	friend struct MMImp;
 
-	static bool                local_dnd_copy;
-	static Ptr<Ctrl>           lastActive;
+	static bool      local_dnd_copy;
+	static Ptr<Ctrl> lastActive;
+	
+	static bool      always_use_bundled_icon;
 	
 	friend void CocoInit(int argc, const char **argv, const char **envptr);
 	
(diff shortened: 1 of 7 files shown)


mirek-fidler:
Well, I have not really checked in macos, but I am afraid that this would break showing the project name in the icon, which I find **extremely** usefull...

klugier:
I reworked it using API badge to provide package name. Here is the documentation https://developer.apple.com/documentation/findersync/fifindersynccontroller/setbadgeimage(_:label:forbadgeidentifier:)?language=objc.

I also attaching screenshot with the new version. I think it works exactly as you want.
Zrzut ekranu 2026-01-10 o 20 26 19

This version also scales well with changes in global system settings.
Zrzut ekranu 2026-01-10 o 21 30 00

mirek-fidler:
Well, merged, but I am not sure I like it...

image

That is on 4K display, I am afraid with 2K it will be useless....

Perhaps we should add an option? (this is just conversational, do not rush to implement it...)

klugier:
My screenshots was taken on a 2K display, so the appearance depends on the dock size. That said, the badge API is designed to ensure the badge remains readable without issues. The same mechanism is used in other apps, for example to display the number of unread messages. If the badge is not readable, that suggests an issue with Apple’s design...

mirek-fidler:
Well, then there is an issue with Apple's design :) I guess this would be fine for number of messages, but the package name is nearly impossible to read (27 inch 4K display). (Note that the screenshot is magnified).

Is it possible there are some properties to the badge (like font size)?

mirek-fidler:
OK, added option to restore previous behaviour (default is "Badge" and bundled icon)

klugier:
Thanks!

PR #341, opened 2026-01-05
UMK: implemented --out-dir optional command line argument.

klugier:
This change introduces new --out-dir command line argument for UMK. The main motivation behind it is to isolate the build to avoid problems. Last week I had an issue when project builds fine with TheIDE, but there were linking problem when building it on Mac. Cleaning cache directory helped to solve this problem. Also, with this change you can specify output directory within repo directory and thanks to that we can force to rebuild the whole project by simply removing build directory.

While 'a' option is useful when you are using umk directly in your terminal, it is less handy when using the build automation tools such as Makefiles or Justfiles.

Usage example:
```
3p/umk/umk.out ./,3p/uppsrc UppTerm 3p/umk/CLANG.bm \
    --hub-dir 3p/hub \
    --out-dir build/cache \
    -br +GUI,SHARED build/upp-term
```

To force rebuild, just run
```
rm -rf build/cache
```

I was also considering --cache-dir, but in the end decided to stay with --out-dor. Please let me know if you have any concerns about naming convention.


uppsrc/umk/umake.cpp
```diff
@@ -237,12 +248,30 @@ CONSOLE_APP_MAIN
...
+			if(!RealizeDirectory(out_dir)) {
+				Puts("Failed to realize output directory \"" + out_dir + "\".");
+				SetExitCode(8);
+				return;
+			}
+			if(!IsFullPath(out_dir)) {
+				out_dir = GetCurrentDirectory() + DIR_SEPS + out_dir;
+				if(!DirectoryExists(out_dir)) {
+					Puts(
+						"Failed to find output directory after relative path noramlization \"" +
```
ismail-yilmaz, line 267: A suggestion: The message seems too verbose (possible confusing) to me, and it contains a typo (should be "normalization").

I think "Couldn't find output directory '$DIR'" would be better here. I mean, this will directly show the normalized path and possibly the problem.

ismail-yilmaz:
Aside from the little recommendation about the error log, this works fine and `--out-dir` makes sense.

ismail-yilmaz (approved):

uppsrc/umk/umake.cpp
mirek-fidler: Will change after merge...

mirek-fidler:
Uhm, now looking at the code, this is pretty confusing:

			if(!RealizeDirectory(out_dir)) {
				Puts("Failed to realize the output directory \"" + out_dir + "\".");
				SetExitCode(8);
				return;
			}
			if(!IsFullPath(out_dir)) {
				out_dir = GetCurrentDirectory() + DIR_SEPS + out_dir;
				if(!DirectoryExists(out_dir)) {
					Puts("Failed to find the output directory \"" + out_dir + "\".");
					SetExitCode(8);
					return;
				}
			}

So we do try to create the potentially relative dir first, then we "normalize" the path check whether it exists? Does not make sense to me...

PR #343, opened 2026-01-07
CtrlCore: Improve minimizebox and maximizebox handling on macos.

klugier:
We are handling sync caption not optimally on macos. To handle it correctly we need to stop using hide and instead of this use more bulletproff styleMas.

Here is the examples.

Before
Zrzut ekranu 2026-01-7 o 21 23 00

After
Zrzut ekranu 2026-01-7 o 21 22 29

Also, the new version is more consistent with how other system apps are handling (Terminal settings - minimize and maximize disabled):
Zrzut ekranu 2026-01-7 o 21 27 13

Zrzut ekranu 2026-01-7 o 21 41 46

On above image these three buttons are always visible, but two of them are disable.

I find another example of our dialog (Print).

Before
Zrzut ekranu 2026-01-7 o 21 36 00

After (more consistent with other apps, however the initial variant might look better)
Zrzut ekranu 2026-01-7 o 21 37 46

Docs https://developer.apple.com/documentation/appkit/nswindow/stylemask-swift.struct?language=objc

### DIFF
--- a/uppsrc/CtrlCore/CocoWin.mm
+++ b/uppsrc/CtrlCore/CocoWin.mm
@@ -290,9 +290,15 @@ void TopWindow::SyncCaption()
 	GuiLock __;
 	if(top) {
 		SyncTitle();
-		NSWindow *window = GetTop()->coco->window;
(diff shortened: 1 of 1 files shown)


mirek-fidler:
Thanks.

PR #347, opened 2026-01-17
CtrlCore: Modernize CocoClip to use modern NSPasteboardTypeFileURL and NSPasteboardTypeURL

klugier:
NSFilenamesPboardType and NSURLPboardType are depricated types from macOS 10.14 and needs to be replaced with there newer equivalents. This PR address this issue by replacing them with NSPasteboardTypeFileURL and NSPasteboardTypeURL. In the end eliminating the warning raised by compiler.

I just noticed that in plain version droping url for example from browser doesn't workg and empty vector is being return. This PR changes this and URL is returned normally. It is returrned in the full form for example:
```
https://examples.com?par1=test
```


mirek-fidler:
Perfect, thanks!

PR #350, opened 2026-01-17
CtrlCore: Implemented drop operation for files in Cocoa.

### LINKED ISSUE #298: Fixing macos clipboard issues
Drag & Drop files doesn't seem to work well on Macos. There are a few issues here:

https://github.com/ultimatepp/ultimatepp/blob/2aa9f2a820ed21c1b627a77ee7f928f13961af5a/uppsrc/CtrlCore/CocoClip.mm#L58

Doesn't work for files, it will use Upp::ToString(type) to convert the type to the string "NSFilenamesPboardType" which should be "files".
I think all the types should be defined here:

```
Upp::String fmt = [type isEqualTo:NSPasteboardTypePNG] ? "png" :
	                  [type isEqualTo:NSPasteboardTypeRTF] ? "rtf" :
					  [type isEqualTo:NSFilenamesPboardType] ? "files" :
...etc...
	                                                          Upp::ToString(type);
```

This will solve the internal dragging of files for an ArrayCtrl, and the commented out code here will work again with a small fix:

https://github.com/ultimatepp/ultimatepp/blob/2aa9f2a820ed21c1b627a77ee7f928f13961af5a/uppsrc/CtrlCore/CocoClip.mm#L378

The:

`"<!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd\">\n"`

needs escaping for the two " there:

`"<!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN\" \"http://www.apple.com/DTDs/PropertyList-1.0.dtd\">\n"`

This doesn't solve dragging from an U++ ArrayCtrl to Finder yet, trying to fix that too.
Cheers.

klugier:
This should fix https://github.com/ultimatepp/ultimatepp/issues/298. DropFiles example works like a charm after this changes. You can drop from and to Finder without any problems.

### DIFF
--- a/uppsrc/CtrlCore/CocoClip.mm
+++ b/uppsrc/CtrlCore/CocoClip.mm
@@ -124,8 +124,6 @@ void AppendClipboard(bool dnd, const char *format, const Value& value, String (*
 {
 	GuiLock __;
 
-	RLOG(String("AppendClipboard: format: ") + format + " value: " + String(value));
-
 	auto& data = ClipboardOwner(dnd)->data;
 
 	for(String fmt : Split(format, ';'))
--- a/uppsrc/CtrlCore/CocoClip.mm
+++ b/uppsrc/CtrlCore/CocoClip.mm
@@ -43,6 +43,7 @@ NSPasteboard *Pasteboard(bool dnd = false)
 	};
 	
 	NSPasteboard *pasteboard = Upp::Pasteboard(dnd);
+	[pasteboard clearContents];
 
 	if([type isEqualTo:NSPasteboardTypeString]) {
 		Upp::String raw = render("text");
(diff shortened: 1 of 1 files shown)


uppsrc/CtrlCore/CocoClip.mm
```diff
@@ -43,6 +43,7 @@ -(void)pasteboard:(NSPasteboard *)sender provideDataForType:(NSString *)type
 	};
 	
 	NSPasteboard *pasteboard = Upp::Pasteboard(dnd);
+	[pasteboard clearContents];
```
klugier, line 46: Should be called everytime we want to save something to pasteboard. Without it the solution doesn't work, so it is safe to call it here.

PR #352, opened 2026-01-18
Ide: Toolbar in the row is now true on macOS

klugier:
Since there is no menubar on macOS and we are alwasy using system wide menu bar. The default behaviour with no toolbar in the row looks strange. So, this PR changes this option. We can also, consider to hide this option from settings...

Toolbar in the row
Zrzut ekranu 2026-01-19 o 00 07 44

No toolbar in the row, too much empty space...
Zrzut ekranu 2026-01-19 o 00 07 58

BTW, We can consider moving the question mark when Toolbar in the Row is active after the current line display indicator.

### DIFF
--- a/uppsrc/ide/ide.h
+++ b/uppsrc/ide/ide.h
@@ -628,7 +628,12 @@ public:
 	byte      default_charset;
 	bool      indent_spaces;
 	bool      show_status_bar;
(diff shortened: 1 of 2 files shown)


mirek-fidler:
Good idea thanks. Let us leave the option there though..

PR #356, opened 2026-01-21
GridCtrl: fix warnings on macOS.

klugier:
Fixing following warnings:
- Remove unused enteres variable in GridDisplay.cpp
- Replaced vsprintf with secure vsnprintf in GridUtils.cpp


mirek-fidler:
I thing the whole CtrlUtils.cpp is anachronism - going to remove it completele after the merge (and switch those LG macros to regular LLOG)

PR #369, opened 2026-02-09
CtrlCore: Fix GTK file drop to Dolphin and Firefox on Wayland

klugier:
It looks like we need to recover slashes after doing UrlEncode. So, instead of having following path:
```
file://%2Fhome%2Fklugier%2Fupp%2Ffr-fr.udc
```
It should be
```
file:///home/klugier/upp/fr-fr.udc
```
After changing the logic drop to Dolphin and Firefox on Wayland works as expected.

I tested also on pure X11 and it works as expected.


mirek-fidler:
I decided to go a bit deeper and implemented EncodeUrlPath. Gemini approves, please check...

PR #370, opened 2026-02-14
Core/SSH: Add tutorial, design docs and FAQ.

ismail-yilmaz:
## Add Documentation for Core/SSH Package

Adds design documentation, FAQ, and practical tutorial to Core/SSH:
All tutorial examples are from the existing SshBasics reference example.


klugier:
I think this is a nice addition overall, but it still needs some formatting polish to match the standard of our existing documentation.

Here are the improvements I would suggest:
- Instead of using the “Default” style, please use the “Text” style for regular paragraphs.
- Add proper code formatting. To do this:
    - Locate the relevant code in TheIDE.
    - Right-click and select Miscellaneous → Copy With Syntax Highlighting.
    - Paste the code into the document.
    - Set the font size to 9.
- Apply the appropriate left margin.

The table of contents formatting is not aligned with similar sections in other documents. There is no need to add extra margin, and the numbering should be included as part of the hyperlink.

I believe this tutorial should follow the same style as the GUI tutorial, which is one of our best examples. Please use it as a reference. If anything is missing from the template, we should update the tutorial template accordingly.

ismail-yilmaz:
Will do.
I think Tutorial template should be updated though. That'll make many things easier.

klugier:
Thanks! I will look at that...

ismail-yilmaz:
@klugier  

I have updated the tutorial and design notes doc and removed faq (it was unnecessary)
Please check.

klugier:
LGTM!

klugier (approved):

PR #371, opened 2026-02-21
Fix shebang line in install script

klugier:
The install script doesn't work on Ubuntu 25.04. It generates following error:
```
./install: 75: bad substituion
```
The reason of that error is that we incorrect sheban in our script that can not force the script execution under bash:
```
##!/usr/bin/env bash
```
It should be:
```
#!/usr/bin/env bash
```

If we won't release soon this should be in hotfix release for 2025.1.


mirek-fidler:
Thanks

PR #373, opened 2026-02-22
Fix MSVC warnings (Feb2026).

klugier:
It looks like we have bunch of new MSVC compilation warnings. This PR address it. We should make sure our new release will be free from MSVC warnings.

### DIFF
--- a/uppsrc/ide/Log.cpp
+++ b/uppsrc/ide/Log.cpp
@@ -213,7 +213,7 @@ void Ide::FindDs(int where, bool all)
 
 	int n = 0;
 	Progress pi;
-	for(String fn : files) {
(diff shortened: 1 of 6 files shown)


uppsrc/ide/Log.cpp
```diff
@@ -213,7 +213,7 @@ void Ide::FindDs(int where, bool all)
 
 	int n = 0;
 	Progress pi;
-	for(String fn : files) {
+	for(const String& fn : files) {
```
mirek-fidler, line 216: What warning does this produce?

mirek-fidler:
Thatks for bringing this up; I have tried to go bit deeper in the last push...

PR #379, opened 2026-03-26
WebSocket SNI fix for SSL connections

ismail-yilmaz:
Currently, secure websocket connections will fail on many servers due to servers requiring SNI information and there is no way to add the info. 

This PR,

- Adds/exposes `SSLServerNameIndication` method to `WebSocket`.
- Updates the `WebSocket` docs.

Please review.


przemkaz:
@mirek-fidler could You take a look at this PR. 

Thank you in advance

PR #383, opened 2026-05-16
umk: Fix crash with incomplete build methods

amozeo:
Fixes umk crash when build method file lacks at least "PATH", "INCLUDE", or "LIB" keys.

Motivation: In my nix dotfiles I have an `buildUppPackage` helper that generates the build method dynamically based on arguments, that did not generate .bm file with those keys by default, causing crashes on build failures on any package using it.

this crash didn't exist on v2025.1.1


mirek-fidler:
Thank you!

PR #403, opened 2026-08-22
Upgrade SQLite3 Multiple Ciphers to v2.5.0 (SQLite v3.53.4)

lukoml:


mirek-fidler:
Note that this might be a very temporary solution - we are moving to vpckg on windows and using host platform libraries for everything on POSIX.

User0755:
Hi! We should retain the existing vendored plugin/sqlite3 (SQLite3 Multiple Ciphers / sqleet) unchanged (although updated like it now is) for backward compatibility with existing encrypted databases, and add a new plugin/sqlcipher using SQLCipher for new encrypted databases. The new plugin should use platform packages where available (Linux distro packages, macOS package manager) and vcpkg on Windows.

Eventually, users should convert their databases to be compatible with sqlcipher so that the vendored code can be dropped in favor of SQLCipher.

BR, Tom

User0755:
I agree. It would be a cleaner solution to have the pure original sqlite3 as plugin/sqlite3 and the derivative versions under their own names somewhere else.

